PR #2, opened 2021-12-06
Preliminary version: supports plugin API.

jsandoe:
There are a number of hacks in this version that probably need to be resolved before incorporation of this PR, but it does work in the sense that the app can be used as a plugin in CODAP or as a standalone page. The hacks:
* I incorporated a .ts library, borrowed from the Story Builder project that had a considerable number of lint errors. To quiet lint, I altered webpack config. There is probably a better way to tell lint to ignore this file.
* I introduced a 2 second delay before render in index.ts. This is to allow CODAP and QP to establish a connection before rendering occurs in diagram.ts. Undoubtedly, there is a better way, but I am assuming that initialization of the MOBX model in diagram.ts is a provisional step, anyway.
* Upgraded from Javascript es5 to es2017. CODAP assumes modern browsers.

### DIFF
--- a/src/components/diagram.tsx
+++ b/src/components/diagram.tsx
@@ -1,5 +1,5 @@
 import { observer } from "mobx-react-lite";
-import { getSnapshot, Instance } from "mobx-state-tree";
+import { getSnapshot, applySnapshot, Instance } from "mobx-state-tree";
 import React, { useState } from "react";
 import ReactFlow, { Edge, Elements, OnConnectFunc, 
   OnEdgeUpdateFunc, OnLoadParams, MiniMap, Controls, isNode } from "react-flow-renderer";
@@ -7,12 +7,13 @@ import { DQRoot } from "../models/dq-models";
 import { DQNode, Operation } from "../models/dq-node";
 import { NodeForm } from "./node-form";
 import { QuantityNode } from "./quantity-node";
+import codapInterface from "../lib/CodapInterface";
 
 let nextId = 0;
 const loadInitialState = () => {
   const url = new URL(window.location.href);
   const urlDiagram = url.searchParams.get("diagram");
-  
+
   // Default diagram
   let diagram = {
     nodes: {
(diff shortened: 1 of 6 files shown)


src/components/diagram.tsx
```diff
@@ -55,6 +56,28 @@ const loadInitialState = () => {
...
+      }
+    },
+    (msg) => {
+      console.log( "CODAP not found: " + msg);
+    }
+);
+
+  codapInterface.on("get", "interactiveState", "",
+      () => {
+        return {success: true, values: getSnapshot(dqRoot)};
```
scytacki, line 78: I think this will not capture the positions of the nodes correctly.
Currently the node positions are not updated in the MST models as a user drags around the node.
If you look at `copyDiagramURL` you can see how I dealt with that.
You can extract out that part of the code so it can be shared here.
However it is relying on the `rfInstance` state variable which won't be available in this global context.

webpack.config.js
```diff
@@ -113,9 +113,9 @@ module.exports = (env, argv) => {
       warningsFilter: /export .* was not found in/
     },
     plugins: [
-      new ESLintPlugin({
-        extensions: ['ts', 'tsx', 'js', 'jsx'],
-      }),
+      // new ESLintPlugin({
+      //   extensions: ['ts', 'tsx', 'js', 'jsx'],
+      // }),
```
scytacki, line 118: Take a look at the .eslintrc.js file. It has an overrides section where you can configure different settings for different file name patterns. I don't know the syntax for basically disabling rule, but there probably is one.

src/index.tsx
```diff
@@ -1,10 +1,14 @@
...
-);
+
+const render = () => {
+    ReactDOM.render(
+        <App />,
+        document.getElementById("app")
+    );
+};
+
+window.setTimeout(render, 2000);
```
scytacki, line 14: What happens without this?
Does QP render then codap initializes, and then the diagram changes to the saved version?
Or is there some bug that happens instead?

Yes the loading of the state in the component is temporary. I wanted to see what needs CODAP would have. And I was also hoping to integrate this into CLUE to see what I had to change to do that.

scytacki (changes requested):
This is looking good.

I'd be fine merging with the 2 second delay.
But the node positions should be handled correctly (unless it already works in some way I don't understand).
And the linting should be turned back on for the rest of the files.

Also I see the automated build and deploy failed do some typescript "linting" errors: https://github.com/concord-consortium/quantity-playground/runs/4434316308?check_suite_focus=true
If you can get those fixed then this would be deployed at:
https://models-resources.concord.org/quantity-playground/branch/qp-as-codap-plugin/

src/components/diagram.tsx
jsandoe: True. Latest changes to node positions are not captured. Will fix.

webpack.config.js
jsandoe: For now, I am adding in the webpack.config class, "exclude: [ 'src/lib', 'node_modules]" to exclude the file from being linted.

src/index.tsx
jsandoe: QP loads the default version then switches to the saved version. I put in the delay to avoid the flash.

scytacki:
I just merged the other PR so I'm going to close this one. Feel free to reopen if it is useful.

PR #4, opened 2022-01-12
Save in codap apply snapshot

jsandoe:
Scott: here is a version that restores my earlier version. It uses applySnapshot to restore the model from CODAP.

To replicate the bug, 
1. run this URL: https://codap.concord.org/app?di=https://models-resources.concord.org/quantity-playground/branch/save-in-codap-apply-snapshot/
2. make some changes, then, in the CODAP hamburger menu select create a copy.
3. CODAP will open in a new tab
4. try to create a new node: the changes made in 2 will vanish with the drop


scytacki:
@jsandoe I played around with it a bit. I think the issue is the id system. If I add 2 new nodes in the original, so there are 5 total. And then make a copy. When I add a new node in the copy, the 4th node disappears and is replaced with the new node. 

I think the "nextId" is causing this. If you recompute nextId when you apply the snapshot I think it will fix this problem.

React Flow should be ignoring the positions of the nodes that it already knows about. So I'm a little surprised the positions of the first 3 nodes get updated when the state from codap is applied. But it seems to be working. 🤷

scytacki (approved):
This looks good to me. 

I'm going to merge it in. I'll work on getting the positions of the nodes updated more frequently in the state tree. That way onSnapshot will fire when the nodes are moved.

PR #5, opened 2022-01-25
update the node positions as reactflow changes them

scytacki:
this has two benefits:
- it simplifies the serialization
- CODAP can just monitor the tree to see changes


jsandoe (approved):
Simpler, cleaner. Looks good to me 👍

PR #6, opened 2022-02-25
Refactor for reusability

kswenson:


src/diagram/models/dq-node.ts
```diff
@@ -267,3 +275,32 @@ export const DQNode = types.model("BasicNode", {
...
         }
     }));
+interface BaseDQNodeSnapshot extends SnapshotIn<typeof BaseDQNode> {}
+
+// importable nodes have more flexible snapshots
+const ImportableDQNode = types.model("ImportableNode", {
+    id: types.identifier,
+    name: types.maybe(types.string),
+    unit: types.maybe(types.string),
+    value: types.maybe(types.maybeNull(types.number)),  // nulls have been encountered in the field
```
scytacki, line 285: We should dig into this more. If the `value` is coming up null in some places then the `unit` and `name` could also show up as null. 

If we do have to handle this, making an additional `ImportableDQNode` seems heavy when we could just do this a couple of lines in the preProcessor function. 

If the point is so that typescript can validate the snapshots we probably should stick with not allowing null in any typescript validated snapshots. And we just have this special null handling until we can figure out where it is coming from.

But I'm going to merge this and we can look at this later.

scytacki (approved):
Looks good to me. The importable dq node is something I want to look into, but I'll do that later. 

I tested it out in CODAP and it seemes to work was well as before. The transform was not applied when a saved CODAP document was re-opened, but I don't think it is worth tracking that down until it becomes a requested feature.

src/diagram/models/dq-node.ts
kswenson: It does seem heavyweight, but the lighter-weight `preprocessSnapshot` is deprecated and this is the non-deprecated replacement. It is true that one could make it appear lighter by casting instead of defining the ImportableDQNode but it seemed like if I was going to be using this new `snapshotProcessor` approach I might as well do it right and get the types right. There might be a way to define the ImportableDQNode in terms of the `BaseDQNode` which would eliminate some of the redundancy but I don't know how far MST model composition/redefinition can be pushed.

PR #9, opened 2022-03-12
new version of mathjs improves simplification at lot

scytacki:
With the improved simplification I could remove the 2 hacks that I added.
I verified the hacks with the tests first before and after upgrading.


src/diagram/models/dq-node.test.ts
```diff
@@ -13,3 +13,22 @@ describe("DQNode", () => {
...
+  it("simplifies '(1/s)^2*m' to 'm/s^2'", () => {
+    const result = tryToSimplify("×", "(1/s)^2", "m");
+    expect(result.unit).toBe("m / s ^ 2");
+  });
+
+  it("simplifies '$/piece / peice/day/worker' to '$/day/worker'", () => {
+    const result = tryToSimplify("×", "$/piece", "piece/day/worker");
+    expect(result.unit).toBe("$ / day / worker");
+  });
+});
```
kswenson, line 34:
```suggestion
});

```

kswenson (approved):
👍 LGTM -- FYI, in the library version I have a separate `package.json` in which I was already using mathjs 10.4.0, along with more recent versions of some of the other packages.

scytacki:
@kswenson is that library version is in a branch that isn't pushed? I thought you might have updated it, but I didn't see a change in main.

kswenson:
> @kswenson is that library version in a branch that isn't pushed? I thought you might have updated it, but I didn't see a change in main.

https://github.com/concord-consortium/quantity-playground/blob/40029cb0e0e9d46a90fdacdfae7fe2ecc6cf1013/diagram-view/package.json#L88

kswenson:
> Ah. I didn't realize the diagram view with rollup had its own separate package.json file.

It needs to because some of the shared dependencies become `peerDependencies`, the build scripts are different, etc.

PR #10, opened 2022-03-21
Shared model support

scytacki:
Lots of stuff here.
- a variable MST model was separated from DQNode. This is now used in CLUE for the variables shared model. DQNode now has a reference to this Variable model.
- a top level AppStore model was added so the diagram can be used outside of CLUE and have the same models.
- the tryToSimplify method was extracted because it is independent. It is likely however that this will be replaced with MathJS's unit support so we probably won't need to run our own simplification in that case.
- the form for editing a "node" was converted to one that edits a variable. We might want to reuse this in the text editor variable dialog. But this form will kind of be going away in the diagram when inline editing is supported.
- the nested-set component was updated to work with variables instead of nodes.
- the initialization logic when run standalone or in CODAP was updated to basically reject any older versions of the serialization format. The format now has a version string that has to match the current version or it will not be loaded.
- many tests were added so there this is now almost full coverage of the variable.ts and dq-node.ts
- the dq-root test uses Jest's snapshot feature to make sure the react flow elements generated from the default digram continue to match.
- the toolbar was updated to only show the "Copy Diagram URL" if the diagram has a getDiagramExport prop. 
- the ids in the react flow elements are now the variable id not the node id. This made generating the react flow elements from a node easier. But it means a new `getNodeFromVariableId(id)` view has to be used instead of `nodes.get(id)`
- Variable and VariableType are exported so they can be used in CLUE.


src/components/app-store.ts
```diff
@@ -0,0 +1,47 @@
...
+  const createVariable = () => {
+    const variable = Variable.create();
+    self.variables.put(variable);
+    return variable;
+  };
+
+  return {
+    removeVariable,
+    createVariable
+  };
```
kswenson, line 41: Out of curiosity, is there a reason to use this form of definition instead of:
```
.actions(self => ({
  createVariable() { ... },
  removeVariable() { ... }
});
```
Is it a limitation imposed by `flow`, for instance?

src/diagram/components/nested-set.tsx
```diff
@@ -1,105 +1,79 @@
...
 };
 
-const NestedSetNode = observer(_NestedSetNode);
-NestedSetNode.displayName = "NestedSetNode";
+const NestedSetVariable = observer(_NestedSetVariable);
+NestedSetVariable.displayName = "NestedSetVariable";
 
 interface IProps {
     // this caused an error for some reason
     // node?: Instance<typeof DQNode>;
```
kswenson, line 63: Seems like this comment could be removed now.
```suggestion
```

src/diagram/components/variable-form.tsx
```diff
@@ -1,63 +1,62 @@
...
-const _NodeForm: React.FC<IProps> = ({ node }) => {
+const _VariableForm: React.FC<IProps> = ({ variable }) => {
   const onValueChange = (evt: any) => {
     // if the value is null or undefined just store undefined
     if (evt.target.value == null) {
-      node.setValue(undefined);
+      variable.setValue(undefined);
     } else {
-      node.setValue(parseFloat(evt.target.value));
+      variable.setValue(parseFloat(evt.target.value));
```
kswenson, line 15: Rather than calling `variable.setValue` twice, I would use something like:
```suggestion
      const newValue = evt.target.value != null ? parseFloat(evt.target.value) : undefined;
      variable.setValue(newValue);
```
and the same for unit, name, and operation below.

src/diagram/models/dq-root.ts
```diff
@@ -1,31 +1,70 @@
...
-        self.nodes.forEach((node) => {
-            elements.push(...node.reactFlowElements);
-        });
-        return elements;
+  getNodeFromVariableId(id: string) {
+    return self.nodeFromVariableMap[id];
+  }
+}))
+.actions(self => ({
+  addNode(newNode: Instance<typeof DQNode>) {
```
kswenson, line 40:
```suggestion
  addNode(newNode: DQNodeType) {
```
For consistency and readability

src/diagram/models/dq-root.ts
```diff
@@ -1,31 +1,70 @@
...
+    self.nodes.put(newNode);
+  },
+  removeNode(node: Instance<typeof DQNode>) {
```
kswenson, line 43:
```suggestion
  removeNode(node: DQNodeType) {
```
For consistency and readability

src/diagram/components/diagram.tsx
```diff
@@ -27,8 +27,9 @@ const nodeTypes = {
 interface IProps {
   dqRoot: DQRootType;
   showNestedSet?: boolean;
+  getDiagramExport?: () => any;
```
kswenson, line 30: Would `unknown` be preferable here?

kswenson (approved):
👍 Just a few questions/suggestions but nothing that need hold things up if they're not convenient.

src/components/app-store.ts
scytacki: Yes, I don't know how to do the flow using the simplified function format.

src/diagram/components/diagram.tsx
scytacki: seems to work, I'll go with it.

src/diagram/components/variable-form.tsx
scytacki: I kind of prefer the original form since it would be easier to debug being on multiple lines. It also helps with code coverage stats. I really wish there was something like an `if` that returned a value. I could do a anonymous function like:
```
const newValue = (function(){
  if (evt.target.value != null) {
    return parseFloat(evt.target.value)
  }
})();
variable.setValue(newValue);
```
But that doesn't look much better to me than just calling setValue twice.

PR #12, opened 2022-03-29
Use mathjs units

scytacki:
This removes the use of the `convert-units` library and instead uses math.js to do the unit conversion.

This has several benefits:
- it means the unit conversion will work for arbitrary expressions that we plan on letting students enter in
- unit conversion works for compound or derived units like `mi/hr` to `m/s`
- multiple aliases for a unit are supported like `m, meter, meters`
- custom units can be created with aliases and conversion equation to other units

This change also unifies how single and multiple inputs are handled. They both now just generate an expression like `a`,`b`, or `a + b`. And then if the output node has a unit then that is expression is combined with `to [unit]` like: `(a + b) to [unit]`.

When expressions are supported, we plan on dropping the editable unit field. And instead the user can change the output unit by using the `to [unit]` syntax in the expression.


scytacki:
Oops I missed a couple of things:
- the size of the built library grew a lot because I was lazy with the imports on math.js
- the diagram-view package.json wasn't updated to remove the convert-units.

src/diagram/models/variable.ts
```diff
@@ -43,117 +43,265 @@ export const Variable = types.model("Variable", {
...
+    } catch (e: any) {
+      // TODO: we should find a way to handle this without throwing an
+      // exception, but I think that will mean changes to MathJS
+      //
+      // We should update MathJS to provide more information here. For other
+      // errors MathJS provides a data property on the error that includes
+      // the character location and more info about the error.
+      if (e.message?.startsWith("Units do not match")) {
+        return {error: "incompatible units"};
+      } else if (e.message?.startsWith("Unexpected type of argument")) {
```
kswenson, line 195: Relying on string contents of error messages seems a bit fragile, but it seems they don't have a numeric error code so it's likely the best we can do. Hopefully, they aren't in the habit of tweaking error message wording on a regular basis. We could write unit tests that verify the error messages if we wanted notification if/when such error messages change but I don't think it need hold up the PR.

src/diagram/models/variable.ts
```diff
@@ -43,117 +43,265 @@ export const Variable = types.model("Variable", {
...
-        const inputAUnit = this.inputA.computedUnit;
+    if (self.numberOfInputs === 0) {
+      // Just return the current unit. 
+      // The current unit might be undefined
+      return {unit: self.unit};
+    }
+
+    const expression = self.expression;
+    if (!expression) {
+      // We should only have a empty expression if there are two inputs and the
```
kswenson, line 217:
```suggestion
      // We should only have an empty expression if there are two inputs and the
```

src/diagram/models/variable.ts
```diff
@@ -43,117 +43,265 @@ export const Variable = types.model("Variable", {
...
+
+    // The input value could be a number so it could be 0 which would be
+    // valid
+    if (self.inputB && inputBMathValue === undefined) {
+      // Because we are using `1` if there is no input value, and we just
+      // return the value if there is no unit, the only reason the
+      // inputAMathValue should be undefined is if there is an error with the
+      // input unit.
+      return {error: "invalid input units"};
+    }
```
kswenson, line 249: There's less duplication here, but as before this could be DRYed up with a `validateInputValueOr1` function. This would pay dividends down the road when there could be arbitrary numbers of inputs.

src/diagram/models/variable.ts
```diff
@@ -43,117 +43,265 @@ export const Variable = types.model("Variable", {
...
-            return {unit: inputAUnit};
-          default:
-            break;
+        if (this.inputA?.computedUnit || this.inputB?.computedUnit) {
+          // If one of the inputs has units and the result is not a Unit
+          // that should mean the units have canceled
+          return {message: "units cancel"};
+        } else {
+          // If neither input has a unit then this is unit-less math so it has
+          // not unit and it isn't an error.
```
kswenson, line 271:
```suggestion
          // no unit and it isn't an error.
```

kswenson (changes requested):
👍 Looks good overall -- just a few suggestions

package.json
```diff
@@ -102,7 +102,7 @@
     "html-webpack-plugin": "^5.5.0",
     "identity-obj-proxy": "^3.0.0",
     "istanbul-lib-coverage": "^3.2.0",
-    "jest": "^27.4.7",
+    "jest": "^27.4.1",
```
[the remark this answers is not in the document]
scytacki: I think I did it because there types aren't updated for 27.4.7
So even though we can technically use the new features, typescript doesn't like it.

The particular feature I found missing was this one:
```
    expect(simp.toJSON()).toMatchObject({
      fixPrefix: false,
      mathjs: "Unit",
      unit: "mi",
      // This is supported in Jest but the types don't allow it
      // value: expect.closeTo(0.00055, 5)
    });
```
I had to work around that by adding `expect(simp.toJSON().value).toBeCloseTo(0.000555, 5);`.

src/diagram/models/variable.ts
scytacki: Yes I agree. I predict we are going to have to work on the error reporting so we can help users figure out what is wrong better. So then we'll work on MathJS. There is an issue in their github repo for this already so I think the change would be welcome.

I had one test for the "Units do not match", but I forgot to add one for the "Unexpected type of argument". It is added now.

src/diagram/models/variable.ts
```diff
@@ -43,117 +43,265 @@ export const Variable = types.model("Variable", {
...
+        return {message: "cannot compute value from inputs"};
+      } else {
+        // If there is no value for this input we cannot compute the output
+        // value. There might be a unit.
+        //
+        // We don't show an explicit error here: The current UI shows NaN for
+        // the value, so this seems like enough of a error message.
+        return {};
+      }
+    }
```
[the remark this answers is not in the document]
scytacki: OK I move this closer to supporting more than 2 inputs and in the process DRY'd up this code.

src/diagram/models/variable.ts
scytacki: This should be better now.

package.json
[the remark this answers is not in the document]
scytacki: So I'm not sure what the right solution is. If we use 27.4.7 it is weird when we can't use features until looking at the version of the types. Or if we leave it at 27.4.1 it isn't obvious why it is out of date. I really wish we could add comments to package.json.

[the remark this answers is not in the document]
kswenson: Two thoughts:
1. Have you tried `27.5.1`?
2. `dependencies-notes.md`

kswenson (approved):
👍 Changes look good

package.json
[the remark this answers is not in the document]
scytacki: As far as I can tell there are no `@types/jest` for 27.5.1 either.
But I found a bit more about this and might try `@jest/types` instead.

PR #13, opened 2022-03-31
handle undefined and empty string better

scytacki:
NaN and +/- Infinity are not supported by JSON. At some points we were saving NaN as the value and as that got serialized
and deserialized it was converted null.

### DIFF
--- a/src/diagram/models/variable.ts
+++ b/src/diagram/models/variable.ts
@@ -185,7 +185,9 @@ export const Variable = types.model("Variable", {
     self.inputB = newInputB;
   },
   setValue(newValue?: number) {
-    self.value = newValue;
+    // NaN and Infinity are not valid JSON, we can probably fix MST to handle this correctly
+    // but for now we just convert it to undefined to be safe.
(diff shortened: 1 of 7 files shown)


src/diagram/models/variable.ts
```diff
@@ -185,7 +185,9 @@ export const Variable = types.model("Variable", {
     self.inputB = newInputB;
   },
   setValue(newValue?: number) {
-    self.value = newValue;
+    // NaN and Infinity are not valid JSON, we can probably fix MST to handle this correctly
+    // but for now we just convert it to undefined to be safe.
+    self.value = newValue !== undefined && isFinite(newValue) ? newValue : undefined;
```
kswenson, line 190: I would use `!= null` for breadth of coverage although the distinction may not matter in this instance.

kswenson (approved):
👍LGTM

src/diagram/models/variable.ts
scytacki: Good point.

PR #14, opened 2022-04-11
expression support

scytacki:
This add expression evaluation. 

The modeling was already using mathjs expressions internally. So with this change the user can directly enter these expressions.

The main extra bit of work was parsing the input variable names from the expression. When the expression is evaluated it uses the input cards or nodes. So the variable names are converted from something like `` `car speed` `` to `input_0`

This doesn't handle the case well when the variable name changes. However when we switch to using chips, we'll have to deal with that. So we can punt that issue for now.


scytacki:
I was going to add more to this, like removing the operation pull down and handling better the case where there are unused inputs which aren't valid.  But I decided to put that off so I could get started on the multiple input modeling change.

eireland (approved):
Looks good

src/diagram/models/mathjs-utils.test.ts
```diff
@@ -0,0 +1,38 @@
...
+      const result = replaceInputNames("`var`", ["foo", "var"]);
+      expect(result).toBe("input_1");
+    });
+
+    it("replaces names without backticks", () => {
+      const result = replaceInputNames("var", ["var"]);
+      expect(result).toBe("input_0");
+    });
+
+    it("replaces names without backticks and mutltiple inputs", () => {
```
kswenson, line 20: "mutltiple" => "multiple"

src/diagram/models/mathjs-utils.test.ts
```diff
@@ -0,0 +1,38 @@
+import { replaceInputNames } from "./mathjs-utils";
+
+describe("mathjs-utils", () => {
+  describe("replaceInputNames", () => {
+    it("replaces backtick variables", () => {
+      const result = replaceInputNames("`var`", ["var"]);
+      expect(result).toBe("input_0");
+    });
+
+    it("replaces backtick variables with multiple inputs", () => {
```
kswenson, line 10: It shouldn't really matter, but since the primary use case for backtick variables is names with spaces there would be some illustrative value in demonstrating that in a test.

PR #15, opened 2022-04-12
181611282 variable card

eireland:
Changes the input mode from the variable form to directly into the variable card.


src/diagram/components/variable-form.tsx
```diff
@@ -1,80 +0,0 @@
...
-const _VariableForm: React.FC<IProps> = ({ variable }) => {
-  const onValueChange = (evt: any) => {
-    // if the value is null or undefined or empty just store undefined
-    if (evt.target.value == null || evt.target.value === "") {
-      variable.setValue(undefined);
-    } else {
-      // If the value is the empty string parseFloat turns that into NaN which
-      // we want to avoid.
-      variable.setValue(parseFloat(evt.target.value));
-    }
```
scytacki, line 18: @eireland this was something I changed that you should move over to the quantity-node.tsx. Basically it is checking for an `""` value as well.

src/diagram/components/quantity-node.tsx
```diff
@@ -1,96 +1,134 @@
...
         id="b"
       />
-      <div style={{padding: "10px"}}>
-        {nodeInside}
+      <div className="variable-info-container">
+        <input className="variable-info name" type="text" placeholder="name" value={variable.name || ""}
+            onMouseDown={e => e.stopPropagation()} onChange={onNameChange} data-testid="variable-name"/>
+        <div className="variable-info-value-unit">
+          <input className="variable-info value" type="number" placeholder="value" onChange={onValueChange} data-testid="variable-value"
+            value={shownValue !== undefined ? shownValue.toString() : ""} onMouseDown={e => e.stopPropagation()} />
```
scytacki, line 91: This will make it so I can't demo the changes the unit code which support. Converting units better. However I can show this in a different branch. So I think this is OK to merge, and we'll deal with it later.

scytacki (changes requested):
I found one issue that we missed when merging. 
I added better error checking in the form to prevent NaN from being saved as the number. 
So you should bring that bit of code over to the new form.

PR #16, opened 2022-04-12
181628119 expression editor UI

eireland:
Adds an expression field to the variable card when there is more than one input.
Adds an icon that opens an expression editor.
Adds an expression editor that lets user enter an expression which is saved and shown in the variable card expression field.


kswenson (approved):
👍 As I'm sure you are aware, this still doesn't handle dragging or clicking outside the dialog, but for what it does it seems okay.

PR #17, opened 2022-04-13
Closes Expression Editor when user clicks out side the dialog.

eireland:
Disables input autofill
Resizes expression editor when the textarea expands.
Escape key closes the expression editor and acts like the Cancel button.
Name field no longer expands when expression field expands.

### DIFF
--- a/src/diagram/components/expression-editor.scss
+++ b/src/diagram/components/expression-editor.scss
@@ -1,5 +1,4 @@
 .expression-editor-dialog {
-  height: 125px;
   width: 510px;
   position: absolute;
   top: 85px;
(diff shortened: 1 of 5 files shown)


eireland:
@scytacki This does not include the merge with expression-support branch.

scytacki (changes requested):
The changes look good. 

Can you add some more test coverage?
At least the escape and enter key hander handlers can be tested with react testing library use-event package:
https://testing-library.com/docs/ecosystem-user-event#special-characters

scytacki (approved):
Looks great to me.

PR #18, opened 2022-04-14
remove operation support, and automatic single input expression

scytacki:
this also switches all of the tests to use expressions


scytacki:
@eireland if you think this looks OK, and you want to merge it, that is fine with me.

Or you could add to it and remove the operation UI too.

eireland (approved):
👍

PR #19, opened 2022-04-21
Prevents node from being deleted when backspace or delete is typed in…

eireland:
… expression editor


src/diagram/components/expression-editor.tsx
```diff
@@ -13,6 +13,14 @@ export const ExpressionEditor: React.FC<IProps> = ({variable, onShowExpressionEd
   const [appliedExpression, setAppliedExpression] = useState(variable.expression);
   const expressionEditorTextAreaRef = useRef<HTMLTextAreaElement | null>(null);
 
+  useEffect(() => {
```
kswenson, line 16: Add a comment about why this is necessary and why it's implemented the way it is.

kswenson (approved):
👍 Looks good -- just needs a comment.

PR #20, opened 2022-04-28
181836284 multiple inputs

eireland:
Changes quantity node to take in multiple inputs.
Changes input lines so they "float" into the node.


src/diagram/models/variable.ts
```diff
@@ -261,7 +261,7 @@ export const Variable = types.model("Variable", {
           return {unit: unitString};
         }
       } else {
-        if (self.inputs[0]?.computedUnit || self.inputs[1]?.computedUnit) {
+        if (nodeInputs[0]?.computedUnit || nodeInputs[1]?.computedUnit) {
```
scytacki, line 264: This should be changed to a loop that goes over all of the inputs. If any of the inputs have a unit then we show the "units cancel" message:

```suggestion
        if (nodeInputs.some(input => input?.computedUnit) {
```

src/diagram/models/variable.ts
```diff
@@ -328,11 +328,8 @@ export const Variable = types.model("Variable", {
   }
 }))
 .actions(self => ({
-  setInputA(newInputA: Instance<IAnyComplexType> | undefined) {
-    self.inputA = newInputA;
-  },
-  setInputB(newInputB: Instance<IAnyComplexType> | undefined) {
-    self.inputB = newInputB;
+  setInput(newInput: Instance<IAnyComplexType> | undefined) {
```
scytacki, line 331: change to `addInput`

src/diagram/models/variable.ts
```diff
@@ -351,5 +348,13 @@ export const Variable = types.model("Variable", {
   setExpression(newExpression?: string) {
     self.expression = newExpression;
   },
+}))
+.actions(self => ({
+  removeInput(input: DQNodeType) {
```
scytacki, line 353: The variable model shouldn't have any dependency on the DQNode. 
So this should be changed to take either a variable instance instead.

src/utils/diagram/floating-edge-util.tsx
```diff
@@ -0,0 +1,69 @@
+import { Position } from "react-flow-renderer/nocss";
```
scytacki, line 1: Because this isn't rendering any react, its extension should be `.ts` instead of `.tsx`.
You should add some tests of it. That would help explain what it does, and keep the coverage up.

scytacki (changes requested):
This looks good.

I left several comments. Some of them are kind of nit picky and nice to have, but others are more important.

I'd say the most important ones are:
- fixing the "units cancel" check
- removing the reference to DQNode from the Variable model
- adding tests of floating-edge-util
- changing `setInput` to `addInput`

src/diagram/models/variable.ts
```diff
@@ -38,20 +39,10 @@ export const Variable = types.model("Variable", {
   //     return sn;
   // })
   value: types.maybe(types.number),
-  inputA: types.maybe(types.safeReference(types.late((): IAnyComplexType => Variable))),
-  inputB: types.maybe(types.safeReference(types.late((): IAnyComplexType => Variable))),
+  inputs: types.array(types.maybe(types.safeReference(types.late((): IAnyComplexType => Variable)))),
```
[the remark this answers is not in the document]
eireland: I get this error when I tried it ```Argument of type 'IVariable' is not assignable to parameter of type 'IAnyComplexType'.
  Type 'IVariable' is missing the following properties from type 'IType<any, any, object>': create, is, validate, describe, and 5 more.```

scytacki (changes requested):
All of the code changes look good.  

But while testing I remembered the issue we saw about deleting arrows. The only way I can delete an arrow is by hovering next to the handle. I think this will be a show stopper for the project using it with students.

Did you look into fixing this? If you can't find a way I should be able to work with you on Monday to work out something.

scytacki (changes requested):
I think this is good, but as I reported in the PT story it still has a few issues.

If the project team thinks it is useful as it is, then I'm fine with merging it.  And we can address those issues later. I'm just flagging it as needing changes so we know we need to wait for that feedback from the project team.

scytacki (approved):
The project team thought this was good enough as is to work with teachers.
The biggest issue is the crash when trying to delete a node with the backspace (delete key on Mac).  That can be done as another PR.

PR #21, opened 2022-04-28
Quantity node UI changes when there is at least one input

eireland:
Removes operation dropdown when there is at least one input
Makes value and unit static html when there is at least one input


src/diagram/components/quantity-node.tsx
```diff
@@ -107,37 +126,34 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
...
-            </div>
-          }
-          { variable.computedUnitMessage &&
-            <div className="error-message">
-                ⓘ {variable.computedUnitMessage}
-            </div>
-          }
-        </div>
+        {variable.numberOfInputs >= 1 ? renderValueUnitUnEditable() : renderValueUnitInput()}
+        {!(variable.numberOfInputs >= 1) &&
```
scytacki, line 130: I don't understand this. If I'm reading it right, it means the error messages will not be shown if there are inputs. But the error messages are most important when there are inputs.

scytacki (changes requested):
I think this will hide the error messages when there are inputs.

src/diagram/components/quantity-node.tsx
```diff
@@ -121,21 +138,18 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
...
-          { // in an enumeration the keys are the names and the values are string or numeric identifier
+          {!(variable.numberOfInputs >= 1) &&
+            <select className="variable-info operation" value={data.node.variable.operation || ""} onChange={onOperationChange}>
+            { // in an enumeration the keys are the names and the values are string or numeric identifier
+            }
+              <option key="none" value="">none</option>
+              {Object.entries(Operation).map(([name, symbol]) =>
+                <option key={name} value={symbol} data-testid={`variable-operation-${name}`}>{symbol}</option>
+              )}
+            </select>
```
scytacki, line 151: This makes more sense than before. However there isn't a need for an operation if there are no inputs. So I think you can just remove this whole select element at this point.

scytacki (changes requested):
It seems like you should just remove the operation from the UI.

scytacki (changes requested):
The code for this looks good now.

I think the UI of the readonly value and readonly unit needs some work, but lets wait to see what the project team thinks. I asked for that feedback in the PT story.

scytacki:
@eireland I changed this so the unit is ignored in the model when there are inputs. This is better than having the component update the model during a render.

scytacki (approved):
With our changes this looks good to me.

PR #22, opened 2022-04-29
181891254 rollup svg import

eireland:
Moves imported svgs into the quantity-node so they no longer need to be imported.


scytacki (changes requested):
I'd really like to support importing SVGs.  So I'll take a crack at this.

scytacki:
As a follow up on this. I tried using the SVGR rollup plugin. However it fails with not very useful error.
```
[!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)
../src/assets/edit.svg (1:0)
```

My hunch is that it is related to this issue:
https://github.com/gregberge/svgr/issues/679#issuecomment-1061071611

The error message shows a relative path, so if the svgr plugin doesn't handle those, then that would explain this.
My guess is that the relative path happens because the rollup config is setting the input to be a relative parent folder. I tried setting this to be an absolute path, but that didn't help. I also tried moving the svgs around so they were in the same folder as the component using them, that didn't help either. 

Because rollup doesn't have verbose logging it is hard to know if I'm on the right track or not. I'm thinking at this point I'd try switching to webpack, we use it to build other libraries so it should be doable. 

But I'll try one more thing first: I'll add an svg and the svgr plugin to slate-editor because its rollup config doesn't refer to a parent folder for the input.

https://react-svgr.com/docs/rollup

kswenson:
> I'm thinking at this point I'd try switching to webpack, we use it to build other libraries so it should be doable.

FYI: you'll probably have to give up the bundling of types.

scytacki:
Looking at what we do in the lara-typescript, we just rollup to bundle the types. So we could take that same approach. That way we get the same build configuration, and we get bundled types.

Because this is going to take more work, @eireland I now think your approach in this PR is the right one for now.  I'm going try it out locally and merge if it works.

scytacki (approved):

PR #23, opened 2022-05-03
fixes the crash when a connected node is deleted

eireland:


scytacki (approved):
looks good

test coverage would be better, but I don't think we have any tests that can select a node.

PR #24, opened 2022-05-11
fix: variable with space is used more than once

scytacki:
[#182140799]


scytacki:
@eireland can you look at this one?

eireland (approved):
👍

PR #25, opened 2022-05-31
handle cycles better

scytacki:
Cycles in the diagram should not crash it now.
It should no longer be possible to connect a card to itself.
This adds a cypress test to verify that self connections are not possible.

Now that there are cypress tests, the code coverage system "discovered" new files to track:
- index.tsx
- CodapInterface.ts

I added a `codecoverage.yml` to ignore the src/lib folder where CodapInterface.ts is located. Covering this file shouldn't be responsibility of the quantity-playground.

Cycles are handled by wrapping the appropriate MobX getters in try/catch blocks. In theory these getters could throw other exceptions, but we haven't seen that in practice yet. I left TODOs in the code for dealing with this better.

Other changes:
- updated user-event library to version 14, in the end this wasn't necessary but it didn't hurt
- setup cypress to build in github actions,
- record data to a cypress dashboard project,
- upload the cypress coverage to codecov.
- change 'no operation' to 'no expression' which is more accurate


src/diagram/models/variable.ts
```diff
@@ -124,9 +124,23 @@ export const Variable = types.model("Variable", {
...
-        // this input should be fine
-        continue;
+      try {
+        if (input.mathValue !== undefined) {
+          // this input should be fine
+          continue;
+        }
+      } catch (error) {
+        // TODO: this can happen when there is a cycle, perhaps it can happen in other cases
+        // MobX errors do not provide programmatically readable info. The are also different
```
kswenson, line 134:
```suggestion
        // MobX errors do not provide programmatically readable info. They are also different
```

src/diagram/models/variable.ts
```diff
@@ -124,9 +124,23 @@ export const Variable = types.model("Variable", {
...
+        // in development mode. The code for this an be seen here:
```
kswenson, line 135:
```suggestion
        // in development mode. The code for this can be seen here: 
```

src/diagram/models/variable.test.ts
```diff
@@ -697,6 +697,23 @@ describe("Variable", () => {
...
+    const container = GenericContainer.create({
+      items: [
+        {id: "inputA", name: "a", inputs: ["inputA"], expression: "b"},
+        {id: "inputB", name: "b", inputs: ["inputB"], expression: "a"},
+      ]
+    });
+    const variable = container.items[1] as VariableType;
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    const warn = jest.spyOn(console, "warn").mockImplementation(() => {});
```
kswenson, line 710:
```suggestion
    const warn = jest.spyOn(console, "warn").mockImplementation(() => null);
```
In a situation like this I generally just use `null` which shouldn't require the `eslint-disable`.

kswenson (approved):
👍 LGTM

PR #26, opened 2022-06-21
182101825 plural unit handling

jsandoe:
Added the Pluralize npm library. It understands many of the irregular plurizations (mouse/mice, radius/radii, etc) .

### DIFF
--- a/src/diagram/models/mathjs-utils.test.ts
+++ b/src/diagram/models/mathjs-utils.test.ts
@@ -3,24 +3,24 @@ import {getMathUnit, replaceInputNames} from "./mathjs-utils";
 describe("mathjs-utils", () => {
   describe("getMathUnit", () => {
     it("get standard unit", () => {
-      const result = getMathUnit(0,'mm');
-      expect(result?.toString()).toBe('0 mm');
+      const result = getMathUnit(0,"mm");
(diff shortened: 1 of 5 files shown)


src/diagram/models/mathjs-utils.test.ts
```diff
@@ -1,6 +1,26 @@
...
+      expect(result?.toString()).toBe("0 cat");
+    });
+    it("plural of existing custom unit is same", () => {
+      const singU = getMathUnit(0, "mouse");
+      const plurU = getMathUnit(0, "mice");
+      expect(singU && plurU?.equals(singU)).toBe(true);
+    });
+    it("plural of existing custom unit is same 2", () => {
+      const singU = getMathUnit(0, "halves");
+      const plurU = getMathUnit(0, "half");
```
kswenson, line 20:
```suggestion
      const singU = getMathUnit(0, "half");
      const plurU = getMathUnit(0, "halves");
```
Aren't these reversed?

jsandoe: Thanks, I made the test to verify you can create the unit from the plural before the singular exists, but I should have fixed the naming.

scytacki (approved):
Looks great to me.

PR #27, opened 2022-07-19
181735790 card color selection

jsandoe:
There is a new button in the lower right. Clicking it exposes (or hides) a color picker.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "nanoid": "^3.3.1",
         "pluralize": "^8.0.0",
         "react": "^17.0.2",
+        "react-color": "^2.19.3",
         "react-dom": "^17.0.2",
         "react-flow-renderer": "^9.7.4"
       },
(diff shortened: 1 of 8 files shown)


jsandoe:
@scytacki, I requested your review, but then realized I should deal with the code coverage issue, above, and on reading the story more carefully, I see I need to color the backgrounds of related "chips", so I don't thing the PR is quite ready. It is not immediately obvious how to withdraw the request.

scytacki:
@jsandoe I don't think you should do the work to color the chips, that would mean digging into the CLUE codebase where all of those chips are implemented. As long as the color is set on the variable model like you did, it will be available to the components that are rendering the chips. If you want to dig into CLUE to make the change, go ahead, but again I think it'd be fine to skip that if you want.

To remove the review you can usually hit the gear icon to the right of reviewers and then toggle any developers you want to remove.

kswenson:
> To remove the review you can usually hit the gear icon to the right of reviewers and then toggle any developers you want to remove.

You can also mark it "Not ready for review" which puts it in "Draft" mode.

jsandoe:
@scytacki , I will skip the Clue changes, then. The listed "chips" include "Expression Editor". Is there a distinct Expression Editor in Clue or does this refer to the one in QP?

scytacki:
The expression editor doesn't display any chips yet, so you can skip that as well.

scytacki (approved):
Looks great to me!

PR #28, opened 2022-07-20
[#181614001] Variable description can be entered by the user.

jsandoe:
* The story describes the description text area as a part of
  the expanded Card. That story (#181613891) has not been
  worked on yet, so this places the edit field in the normal
  sized card.
  Presumably, the element will be simply migrated to the expanded
  card when it exists.
* The description field can easily run below the bottom of the
  card. I thought about expanding the card in response, but this
  effort would likely be undone by the expanded card story, so
  this issue is left unaddressed.
* Other editable fields in the card are right aligned. It felt
  wrong typing into a right aligned text area, so I left it left-
  aligned. Can be easily changed, though.
* The description field grows to four lines, and then becomes
  scrollable.

### DIFF
--- a/src/diagram/components/quantity-node.scss
+++ b/src/diagram/components/quantity-node.scss
@@ -154,4 +154,16 @@ $gray-mid: #bcbcbc;
     padding: 2px 6px;
     border: 0 solid #999;
   }
+
+  .variable-description-area {
+    resize: none;
+    width: 100%;
+    text-align-last:left!important;
+    border-radius: 5px;
+    border: none;
+    margin: 3px;
+    padding-left: .5em;
+    padding-right: .5em;
+    overscroll-behavior-y: auto;
+  }
 }
--- a/src/diagram/components/quantity-node.scss
+++ b/src/diagram/components/quantity-node.scss
@@ -164,6 +164,18 @@ $gray-mid: #bcbcbc;
     margin: 3px;
     padding-left: .5em;
     padding-right: .5em;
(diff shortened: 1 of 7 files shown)


src/diagram/components/quantity-node.scss
```diff
@@ -154,4 +154,16 @@ $gray-mid: #bcbcbc;
     padding: 2px 6px;
     border: 0 solid #999;
   }
+
+  .variable-description-area {
+    resize: none;
+    width: 100%;
+    text-align-last:left!important;
```
kswenson, line 161:
```suggestion
    text-align-last: left !important;
```
I think we can spare the additional spaces. 😉

src/diagram/components/quantity-node.tsx
```diff
@@ -144,6 +153,11 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
           </div>
         }
         {hasExpression ? renderValueUnitUnEditable() : renderValueUnitInput()}
+        <div className="variable-info-row">
+          <TextareaAutosize className="variable-description-area"
+                            onChange={onDescriptionChange} minRows={1}
+                            maxRows={4}  placeholder={"description"} data-testid={"variable-description"}/>
```
kswenson, line 159:
```suggestion
                            maxRows={4} placeholder={"description"} data-testid={"variable-description"}/>
```

kswenson (approved):
👍  I didn't try to run it locally, but code looks good!

PR #30, opened 2022-10-19
Edit Variable Dialog

tealefristoe:
This PR adds a dialog to edit variables, which can be imported and used by other applications (like CLUE).

### DIFF
--- a/src/diagram/components/diagram.tsx
+++ b/src/diagram/components/diagram.tsx
@@ -1,5 +1,5 @@
 import { observer } from "mobx-react-lite";
-import React, { useRef, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import ReactFlow, { Edge, Elements, OnConnectFunc, isEdge,
   OnEdgeUpdateFunc, MiniMap, Controls, ReactFlowProvider, FlowTransform, OnConnectStartFunc, OnConnectEndFunc } from "react-flow-renderer/nocss";
 import { DQRootType } from "../models/dq-root";
@@ -28,13 +28,19 @@ const edgeTypes = {
 
 interface IProps {
   dqRoot: DQRootType;
+  showEditVariableDialog?: () => void;
   showNestedSet?: boolean;
   getDiagramExport?: () => unknown;
 }
-export const _Diagram = ({ dqRoot, getDiagramExport }: IProps) => {
+export const _Diagram = ({ dqRoot, getDiagramExport, showEditVariableDialog }: IProps) => {
   const reactFlowWrapper = useRef<any>(null);
-  const [ ,setSelectedNode] = useState<DQNodeType | undefined>();
+  const [selectedNode ,setSelectedNode] = useState<DQNodeType | undefined>();
   const [rfInstance, setRfInstance] = useState<any>();
+
+  // Keep the model's selected node up to date with ReactFlow's
+  useEffect(() => {
+    dqRoot.setSelectedNode(selectedNode);
+  }, [dqRoot, selectedNode]);
 
   const handleChangeFlowTransform = (transform?: FlowTransform) => {
     transform && dqRoot.setTransform(transform);
@@ -162,7 +168,7 @@ export const _Diagram = ({ dqRoot, getDiagramExport }: IProps) => {
           onMoveEnd={handleChangeFlowTransform}>
           <MiniMap/>
           <Controls />
-          <ToolBar {...{getDiagramExport}}/>
+          <ToolBar {...{getDiagramExport, showEditVariableDialog}}/>
         </ReactFlow>
       </ReactFlowProvider>
     </div>
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -1,6 +1,8 @@
 import { applySnapshot, getSnapshot, onSnapshot } from "mobx-state-tree";
-import React from "react";
+import React, { useState } from "react";
 import { Diagram } from "../diagram/components/diagram";
+import { DQNodeType } from "../diagram/models/dq-node";
+import { EditVariableDialog, updateVariable } from "../diagram/components/edit-variable-dialog";
 import { AppStore } from "./app-store";
 import codapInterface from "../lib/CodapInterface";
 import defaultDiagram from "./default-diagram";
@@ -83,9 +85,19 @@ const getDiagramExport = () => {
 };
 
 export const App = () => {
+  const [showEditVariableDialog, setShowEditVariableDialog] = useState(true);
+  const nodes: DQNodeType[] = [];
+  appStore.diagram.nodes.forEach((node) => nodes.push(node));
+
   return (
     <div className="app">
       <Diagram dqRoot={appStore.diagram} {...{showNestedSet, getDiagramExport}} />
+      {showEditVariableDialog && nodes.length > 0 &&
+        <EditVariableDialog
+          onClose={() => setShowEditVariableDialog(false)}
+          onSave={updateVariable}
+          variable={nodes[0].variable}
+        />}
     </div>
   );
 };
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -1,7 +1,6 @@
 import { applySnapshot, getSnapshot, onSnapshot } from "mobx-state-tree";
 import React, { useState } from "react";
 import { Diagram } from "../diagram/components/diagram";
-import { DQNodeType } from "../diagram/models/dq-node";
 import { EditVariableDialog, updateVariable } from "../diagram/components/edit-variable-dialog";
 import { AppStore } from "./app-store";
 import codapInterface from "../lib/CodapInterface";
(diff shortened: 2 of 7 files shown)


src/diagram/components/diagram.tsx
```diff
@@ -28,14 +28,20 @@ const edgeTypes = {
...
   const reactFlowWrapper = useRef<any>(null);
-  const [ ,setSelectedNode] = useState<DQNodeType | undefined>();
+  const [selectedNode, setSelectedNode] = useState<DQNodeType | undefined>();
   const [rfInstance, setRfInstance] = useState<any>();
 
+  // Keep the model's selected node up to date with ReactFlow's
+  useEffect(() => {
+    dqRoot.setSelectedNode(selectedNode);
+  }, [dqRoot, selectedNode]);
+
```
scytacki, line 44: Rather than this approach, I'd suggest just calling `dqRoot.setSelectedNode(selectedNode);` in the onSelectionChange handler.  This will make the selection update faster instead of having to wait for a render loop to happen.
Then I think you can get rid of the setSelectedNode state all together. 

I remember now setSelectedNode was here. We used to have a form (like the one you are adding here). And it was rendered based on the selectedNode state.

src/diagram/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+          maxLength={maxCharacters}
+          value={value}
+          onChange={e => setValue(e.target.value)}
+          cols={textarea.cols}
+          rows={textarea.rows}
+        />
+        : <input
+          className="dialog-input dialog-text"
+          id={label}
+          type="text"
```
scytacki, line 37: The type of the value input is number. This allows it handle decimals kind of well. It prevents non-numbers from being entered. And the round trip from onChange back to value works even with a MST property that is a number instead of a string.
But it does have problems with some decimals for some reason.
It seems like this approach here is working, but I haven't groked all of the code yet.

src/diagram/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+interface INormalDialogRow {
+  label: string;
+  maxCharacters?: number;
+  setValue: (value: string) => void;
+  textarea?: ITextArea; // Include { cols, rows } to render a textarea instead of text input
+  value: string;
+  width?: number; // Width of text input
+}
+export const NormalDialogRow = ({ label, maxCharacters, setValue, textarea, value, width }: INormalDialogRow) => {
```
scytacki, line 17: It'd be helpful to rename this file to dialog-row or normal-dialog-row.

src/diagram/components/dialog/edit-variable-dialog.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface IEditVariableDialogContent {
+  name: string;
+  setName: React.Dispatch<React.SetStateAction<string>>;
+  notes: string;
+  setNotes: React.Dispatch<React.SetStateAction<string>>;
+  value: string;
+  setValue: React.Dispatch<React.SetStateAction<string>>;
+  unit: string;
+  setUnit: React.Dispatch<React.SetStateAction<string>>;
+}
```
scytacki, line 20: I think it would be better to just pass a Variable model into this component. Then you don't have to duplicate all of fields in the interface here.

It is still probably best to explicitly define the row components for each field since doing that automatically usually is more trouble than it is worth.

src/diagram/components/dialog/edit-variable-dialog.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface IEditVariableDialog {
+  onClose: () => void;
+  onSave: (updates: IUpdateVariable) => void;
+  variable: VariableType;
+}
+export const EditVariableDialog = ({ onClose, onSave, variable }: IEditVariableDialog) => {
+  const [name, setName] = useState(variable.name || "");
+  const [notes, setNotes] = useState(variable.description || "");
+  const [value, setValue] = useState(variable.value?.toString() || "");
+  const [unit, setUnit] = useState(variable.unit || "");
```
scytacki, line 71: Related to above, I'd try letting the parent component pass in a copy of the Variable object. Then onSave the parent could copy the contents of this variable to the real variable. 
This way you don't have to maintain all of this state.

This approach will bring back the decimal problems with the variable.value, but as long as we have to solve those in the quantity-node component (card) we might as well solve them here.

src/diagram/components/quantity-node.tsx
```diff
@@ -154,7 +154,7 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
         }
         {hasExpression ? renderValueUnitUnEditable() : renderValueUnitInput()}
         <div className="variable-info-row">
-          <TextareaAutosize className="variable-description-area"
+          <TextareaAutosize className="variable-description-area" value={variable.description || ""}
```
scytacki, line 157: Nice catch.

scytacki (changes requested):
This looks good. I've got big requests with more details in the comments:
- get rid of setSelectedNode state in diagram.tsx
- have the dialog work with a copy of the variable directly instead of mirroring each property of the variable with react state variables.

This change will likely require you to support `<input type="number"...` fields in the NormalDialogRow component.

package.json
```diff
@@ -125,6 +126,7 @@
   },
   "dependencies": {
     "iframe-phone": "^1.3.1",
+    "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
tealefristoe: I'm only including lodash for cloneDeep, and I'm sure there's a lighter weight alternative out there. Should I look for one?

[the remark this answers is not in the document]
tealefristoe: Actually I was just able to get rid of cloneDeep, so I removed lodash from the project.

src/diagram/components/dialog/dialog.tsx
tealefristoe: I ended up keeping the input type as text for the value, but added a little validation. It gives some color feedback if an illegal number is entered, and uses the last legal number when an invalid number is present while saving. I was able to get the type="number" system working, but it was having the same decimal issues that are described in this story: #182275661. What I've written here isn't perfect, but it'll give folks an alternative to consider when it comes time to dealing with that story.

scytacki (approved):
This looks great to me. 

The use of text state for the number input field is good. This is what I figured we'd have to do for that other story. 

Perhaps we can extract that as a `NumberInput` component which can be used by the Card and the DialogRow. I've looked for thirdparty generic components that handle this, but haven't found one yet. Rolling our own seems simple enough.

package.json
```diff
@@ -82,6 +82,7 @@
     "@testing-library/react-hooks": "^7.0.2",
     "@testing-library/user-event": "^14.2.0",
     "@types/jest": "^27.4.1",
+    "@types/lodash": "^4.14.186",
```
kswenson, line 85: Why are we adding this as a dependency when there doesn't appear to be any `lodash` dependency, e.g. `lodash` is not listed as a dependency?

src/diagram/components/dialog/dialog-row.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const classes = classnames("dialog-input", { invalid }, textarea ? "dialog-textarea" : "dialog-text");
+  return (
+    <div className="dialog-row">
+      <label className="dialog-label" htmlFor={label}>
+        {label}
+      </label>
+      { textarea
+        ? <textarea
+          className={classes}
+          id={label}
```
kswenson, line 30: Given that `id`s are required to be unique throughout the page, I worry that using the label as the `id` could result in `id` conflicts, given that there's nothing to prevent other items on the page from having an `id` like `Name` or `Notes`.

src/diagram/components/dialog/dialog-row.tsx
```diff
@@ -0,0 +1,50 @@
...
+          maxLength={maxCharacters}
+          value={value}
+          onChange={e => setValue(e.target.value)}
+          cols={textarea.cols}
+          rows={textarea.rows}
+        />
+        : <input
+          className={classes}
+          id={label}
```
kswenson, line 39: Same concern as above.

src/diagram/components/dialog/dialog.scss
```diff
@@ -0,0 +1,52 @@
...
+  text-align: right;
+  font-size: 12px;
+  font-weight: normal;
+}
+
+.dialog-input {
+  border-radius: 5px;
+  border: 1.5px solid #949494;
+  padding: 3.5px 6px;
+  font-family: PTSans-Narrow, sans-serif;
```
kswenson, line 35: I'm curious about this font choice (https://fonts.google.com/specimen/PT+Sans+Narrow). Is it guaranteed that it will be available at runtime?

src/diagram/components/dialog/dialog-row.tsx
tealefristoe: I would also be worried about this, but who else would be crazy enough to use a generic id like Name or Notes? ;)

src/diagram/components/dialog/dialog.scss
tealefristoe: I really don't know the answer to this one. I just used the font Michael had in the spec.

kswenson: If you look at the `index.html` in CLUE, we have lines like these:
```html
    <link href="https://fonts.googleapis.com/css?family=Lato:400,900&display=swap" rel="stylesheet">
    <link href="https://fonts.googleapis.com/css?family=Noto+Sans+Math&display=swap" rel="stylesheet">
    <link href="https://fonts.googleapis.com/css?family=Noto+Sans+Symbols+2&display=swap" rel="stylesheet">
```
According to https://fonts.google.com/specimen/PT+Sans+Narrow, this should be the following for this font:
```html
<link rel="preconnect" href="https://fonts.googleapis.com">
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
<link href="https://fonts.googleapis.com/css2?family=PT+Sans+Narrow:wght@400;700&display=swap" rel="stylesheet">
```
and that same page says the CSS should be:
```css
font-family: 'PT Sans Narrow', sans-serif;
```

tealefristoe: This is very useful, thank you. To not hold up this story any longer, I'm going to move ahead without addressing this, but I'll take care of it next time I need to modify quantity playground, which should be very soon.

I notice that quantity playground has an `index.html`, but diagram view doesn't. Should the links you list above be added to quantity playground `index.html`, to `src/diagram/index.ts`, to CLUE's `index.html`, or something else?

kswenson: I'll let @scytacki answer that one as he's more familiar with the use of standalone quantity playground.

PR #31, opened 2022-10-21
Update yalc local testing

tealefristoe:
This PR updates rollout.config to work with the latest versions of yalc and rollout. It also adds instructions to the readme for testing changes in CLUE without deploying a new version of quantity playground.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,10 @@ Alternately, you can run secure without certificates in Chrome:
 2. Change flag from disabled to enabled
 3. Run `npm run start:secure:no-certs` to run `webpack-dev-server` in development mode with hot module replacement
 
+### Testing in CLUE and other projects
+
+The diagram-view is used in CLUE in the diagram tile, and possibly other projects as well. As you're making changes to this library, it can be helpful to test those changes within client projects without deploying. For more on this, see README.md in the diagram-view directory.
+
 ### Building
 
 If you want to build a local version run `npm build`, it will create the files in the `dist` folder.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -62,6 +62,19 @@ To deploy a production release:
     3. Type in the tag name you want to release for example `v1.2.3`.  (Note this won't work until the PR has been merged to master)
     4. Click the `Run Workflow` button.
 
+### Testing in CLUE
+
+The diagram-view is used in CLUE in the diagram tile, and it can be useful to test changes made in Quantity Playground in CLUE without deploying updates. To do so, it's best to use `yalc`.
+
+1. Install Yalc by running `sudo npm install --global yalc`.
+2. Install Rollup by running `sudo npm install --global rollup`.
```
kswenson, line 70: I generally don't install `yalc` globally, preferring to just use `npx yalc ...` as needed. For tools like `rollup`, generally local install is preferred. Is there a reason you're recommending global install?

README.md
```diff
@@ -62,6 +62,19 @@ To deploy a production release:
...
+3. In the `diagram-view` folder, run `yalc publish`.
+4. In your `collaborative-learning` directory, run `yalc add @concord-consortium/diagram-view`.
+
+After you're happy with your changes and have deployed a new version of Quantity Playground, make sure you revert to using the published version by running `yalc remove @concord-consortium/diagram-view` in your `collaborative-learning` directory.
+
+For more information, see `README.md` in the `diagram-view` directory.
```
scytacki, line 76: To reduce redundancy, I think there should just be a small amount of text in the top level README that refers to the README in the diagram-view directory. If there is info in your text that isn't the diagram-view README, it should be added. Feel free to revise that README to make it more clear.

scytacki (changes requested):
Code changes look good. I'd just suggest refactoring the readme doc changes.

README.md
tealefristoe: I was following the advice from these sites:
https://github.com/wclr/yalc
https://rollupjs.org/guide/en/

Installing globally is how I was able to get this to work, so that's what I included in my writeup. However I don't have strong feelings about it and will change these lines if you think there's a better way to do it.

kswenson: For libraries like `webpack` and `rollup`, it is common for different versions to be used in different projects, so local installation is generally preferred. For `yalc`, it's not actually a dependency of the project so the same concern doesn't apply, but personally I still try to avoid global npm installs and so prefer to use `npx yalc ...` instead.

tealefristoe: Alright that's fine with me. What should I replace the lines I wrote with? And will someone have to install yalc separately in both projects using npx?

kswenson: The `rollup` dependency should be listed in the `package.json` with the other dependencies and should therefore be installed via `npm install` along with the other dependencies. As for `yalc`, `npx` supports executing packages without installing them, so no installation of `yalc` (global or local) should be necessary to run `npx yalc ...`. In short, neither of these lines should be necessary.

scytacki (approved):
looks good to me

PR #34, opened 2022-11-04
182275661 validation

tealefristoe:
- Unify validation between variable cards and dialogs

### DIFF
--- a/src/diagram/components/quantity-node.tsx
+++ b/src/diagram/components/quantity-node.tsx
@@ -1,12 +1,15 @@
+import classnames from "classnames";
 import { observer } from "mobx-react-lite";
 import { isAlive } from "mobx-state-tree";
 import React, { useState } from "react";
 import { Handle, Position } from "react-flow-renderer/nocss";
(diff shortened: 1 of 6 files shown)


src/diagram/components/quantity-node.tsx
```diff
@@ -1,12 +1,15 @@
+import classnames from "classnames";
```
kswenson, line 1:
```suggestion
import classNames from "classnames";
```
We generally camel-case the import

src/diagram/utils/validate.ts
```diff
@@ -1,2 +1,6 @@
 export const kMaxNameCharacters = 30;
 export const kMaxNotesCharacters = 255;
+
+export const validNumber = (v: string) => !isNaN(+v);
```
kswenson, line 4: I would name this
```suggestion
export const isValidNumber = (v: string) => !isNaN(+v);
```
to make it clear that it's returning a boolean rather than converting to a valid number, for instance.

kswenson (approved):
👍 Looks good -- one trivial naming suggestion

PR #35, opened 2022-11-08
Variable Lists

tealefristoe:
- Move `VariableChip` from CLUE and add new features:
  - Allow an onClick function
  - Make selectable
  - Allow a className
  - Optionally omit the value and unit if the variable has a name (name only chip)
- Introduce `VariableChipList` which renders multiple chips.
  - Allows for a list of selected variables
  - This will be used in various dialogs in CLUE


scytacki (changes requested):
This looks good.

Could you add someway to show the chip list within the stand alone quantity-playground? Otherwise this will look like dead code if the quantity playground is being worked on in isolation. 

There used to be some URL params which would turn on various views. I'm not sure if those are still around, but that would be one option for displaying the variable list.

scytacki (approved):
Looks great!

src/diagram/components/ui/variable-chip-list.test.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+// import { applySnapshot } from "mobx-state-tree";
```
picklesrus, line 2: minor: remove unsued import?

src/diagram/components/ui/variable-chip.scss
```diff
@@ -0,0 +1,35 @@
...
+    outline: 2px solid vars.$select-blue;
+  }
+
+  // So selection matches text selection in slate
+  &.slate-selected {
+    outline: 2px solid lightskyblue;
+  }
+}
+
+// .variable-chip-highlight {
```
picklesrus, line 33: remove?

src/diagram/components/ui/variable-chip.tsx
```diff
@@ -0,0 +1,56 @@
...
+  selected?: boolean;
+  variable: VariableType;
+}
+
+export const VariableChip = observer(({ className, nameOnly, onClick, selected, variable }: IVariableChip) => {
+  const name = variable.name;
+  const value = variable.computedValue;
+  const valueString = variable.computedValueWithSignificantDigits;
+  const unit = variable.computedUnit || "";
+  const showValue = value !== undefined && !isNaN(value);
```
picklesrus, line 24: isNaN(undefined) returns true (I think) so maybe you don't need the first part?

picklesrus (approved):
LGTM. I made a few very minor comments.

src/diagram/components/ui/variable-chip.tsx
tealefristoe: I tried removing the first part, and was surprised to get a type error. Apparently `isNaN()` only accepts numbers, which `NaN` technically is. Seems like it would be a lot more useful if it accepted `any` and just returned false for anything that isn't a number ¯\_(ツ)_/¯

picklesrus: Interesting! Thanks for explaining!

PR #36, opened 2022-11-14
183640291 Separate Variables from Diagram View

tealefristoe:
- Adds highlighting to selected variable card
- Removes the X button from variable cards and adds a delete button to the toolbar
- Keeps variables around in the store even when the associated variable card is deleted
- Adds a dialog to view and delete unused variables
- Adds a hook to streamline selecting multiple variables
- Allows custom styling of variable chip lists and adds disabled style for variable chips
- Changes how number inputs handle invalid values
  - When an invalid value is specified, the value is saved as `undefined` rather than the last valid value entered
  - `undefined` can be explicitly specified by entering ""


src/diagram/components/quantity-node.tsx
```diff
@@ -126,8 +127,13 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
   const targetNodeHandleStyle = {height: nodeHeight, width: nodeWidth, left: "1px", opacity: 0, borderRadius: 0};
   const sourceHandleStyle = {border: "1px solid white", borderRadius: "6px", width: "12px", height: "12px", background: "#bcbcbc"};
 
+  const classnames = classNames(
+    "node",
+    hasExpression && "expression-shown",
+    data.dqRoot.selectedNode === data.node && "selected"
+  );
```
scytacki, line 134: This is subjective, but using classNames' hash argument makes this a little more clear to me. I think this is approach is uses in some parts of CLUE.
```suggestion
  const classnames = classNames({
    node: true,
    "expression-shown": hasExpression,
    selected: data.dqRoot.selectedNode === data.node
  });
```

src/diagram/components/ui/variable-chip-list-container.tsx
```diff
@@ -1,28 +1,19 @@
...
-    selectedVariables.forEach(v => {
-      if (v === variable) {
-        found = true;
-      } else {
-        newSV.push(v);
-      }
-    });
-    if (!found) newSV.push(variable);
-    setSelectedVariables(newSV);
+    toggleVariable(variable);
```
scytacki, line 16: Maybe it is better to just drop `onClick` and just pass toggleVariable directly to the component.

scytacki (approved):
This seems good to me.
I just left a couple of minor comments.

scytacki (approved):
This looks good to me.

I just found one thing that seemed odd in a test.

src/diagram/components/quantity-node.test.tsx
```diff
@@ -77,9 +76,8 @@ describe("Quantity Node", () => {
     expect(variable.description).toBe("a\ndescription");
 
     //verify entering a non-number does not change the value
-    const oldValue = variable.value;
     await userEvent.type(valueTextBox, "letter");
-    expect(variable.value).toBe(oldValue);
+    expect(variable.value).toBe(45);
```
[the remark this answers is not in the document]
tealefristoe: It's looking at the variable, which won't update until the input loses focus, so it's correct, if not particularly informative. I'll add the `oldValue` variable back in so it's more clear what it's checking.

PR #40, opened 2022-11-30
M2Studio UX - Format Var Name

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183570281

[#183570281]

This change completes the requirements for the story by ensuring that the variable name field on the card only allows up to 27 characters. (Note that the PT story specifies 30 characters, but the specs say 27.)

Other requested features already existed were or were completed in an earlier PR.


tealefristoe (approved):
Short & sweet

PR #41, opened 2022-11-30
M2Studio UX - Set spellcheck="false" attribute

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183914399

[#183914399]

Disables spellcheck for variable name, value, unit, and expression fields. The notes/description field should still have spellcheck enabled.


tealefristoe (approved):
Looks good

PR #42, opened 2022-12-01
M2Studio UX - Format Value and Unit Field

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183570351

[#183570351]

Value and Unit fields are now expandable when the entered values are too long to be contained within the default field size. This feature required changing each of the fields from an input of type text to a textarea.

### DIFF
--- a/src/diagram/components/quantity-node.tsx
+++ b/src/diagram/components/quantity-node.tsx
@@ -5,15 +5,15 @@ import { Handle, Position } from "react-flow-renderer/nocss";
 import TextareaAutosize from "react-textarea-autosize";
 import classNames from "classnames";
 import { ColorEditor } from "./color-editor";
-import { NumberInput } from "./ui/number-input";
 import { DQNodeType } from "../models/dq-node";
 import { DQRootType } from "../models/dq-root";
-import { kMaxNameCharacters, kMaxNotesCharacters, processName, isValidNumber } from "../utils/validate";
+import { kMaxNameCharacters, kMaxNotesCharacters, processName } from "../utils/validate";
 import { IconColorMenu } from "./icon-color-menu";
 import { IconExpand } from "./icon-expand";
 import { IconWarning } from "./icon-warning";
 
 import "./quantity-node.scss";
+import { ExpandableInput } from "./ui/expandable-input";
 
 interface IProps {
   data: {node: DQNodeType, dqRoot: DQRootType};
@@ -22,13 +22,17 @@ interface IProps {
 
 const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
   const variable = data.node.variable;
+  const kDefaultExpandLength = 10;
+  const kExpressionExpandLength = 18;
 
   const [showColorEditor, setShowColorEditor] = useState(false);
   const [showDescription, setShowDescription] = useState(false);
-  const [showExpression, setShowExpression] = useState(false);
+  const [showFullExpression, setShowFullExpression] = useState(false);
   const [showFullValue, setShowFullValue] = useState(false);
   const [showFullUnits, setShowFullUnits] = useState(false);
+  const [hasLongExpression, setHasLongExpression] = useState(variable.expression && variable.expression.length > 20);
   const [hasLongValue, setHasLongValue] = useState(variable?.value && variable.value.toString().length > 10);
+  const [hasLongUnits, setHasLongUnits] = useState(variable?.unit && variable.unit.length > 10);
 
   // When the node is removed from MST, this component gets
   // re-rendered for some reason, so we check here to make sure we
@@ -38,8 +42,6 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
   }
 
   const hasExpression = !!(variable.numberOfInputs > 0);
-  const hasLongExpression = !!(variable.expression && variable.expression.length > 20);
-  const hasLongUnits = !!(variable.unit && variable.unit.length > 10);
   const hasError = !!(variable.computedValueError || variable.computedUnitError || variable.computedUnitMessage);
   const shownValue = hasExpression ? variable.computedValue?.toString() || "" : variable.value;
   const shownUnit = hasExpression ? variable.computedUnit : variable.unit;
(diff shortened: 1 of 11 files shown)


tealefristoe (changes requested):
This looks good, but there are a few big questions I have and opportunities for refactoring I see. Unfortunately I have an early meeting this morning so I can't elaborate more now, but I wanted to get the big feedback to you sooner rather than later. Later today I'll try to send more feedback.
- First, and probably biggest, I see a lot of duplicated code between the various input fields that now are expandable. How much of this could be consolidated into reusable components? At the very least, the expand buttons could be a reused component. But it seems likely that the entire package, including input field (textarea or NumberInput), expand button, and container div, could be a reusable component. This would also have the benefit of breaking up `quantity-node.scss`, which is starting to get pretty unwieldy.
- More feedback later today...

emcelroy:
@tealefristoe Good point, thanks. Yeah, `quantity-node.scss` is pretty unwieldy. I actually tried to make the CSS more concise there, but ran into issues. Breaking some things into reusable components is very likely a better approach. I'll start working on that.

src/diagram/components/quantity-node.tsx
```diff
@@ -5,27 +5,34 @@ import { Handle, Position } from "react-flow-renderer/nocss";
...
   const [showDescription, setShowDescription] = useState(false);
-  const [showExpression, setShowExpression] = useState(false);
-
-  const variable = data.node.variable;
+  const [showFullExpression, setShowFullExpression] = useState(false);
+  const [showFullValue, setShowFullValue] = useState(false);
+  const [showFullUnits, setShowFullUnits] = useState(false);
+  const [hasLongExpression, setHasLongExpression] = useState(variable.expression && variable.expression.length > 20);
+  const [hasLongValue, setHasLongValue] = useState(variable?.value && variable.value.toString().length > 10);
+  const [hasLongUnits, setHasLongUnits] = useState(variable?.unit && variable.unit.length > 10);
```
tealefristoe, line 35: Can these be turned into local variables rather than state? This would simplify things, plus avoid duplicating the code in two places (see lines 70 and 76).

src/diagram/components/quantity-node.tsx
```diff
@@ -5,27 +5,34 @@ import { Handle, Position } from "react-flow-renderer/nocss";
...
+  const kExpressionExpandLength = 18;
+
   const [showColorEditor, setShowColorEditor] = useState(false);
   const [showDescription, setShowDescription] = useState(false);
-  const [showExpression, setShowExpression] = useState(false);
-
-  const variable = data.node.variable;
+  const [showFullExpression, setShowFullExpression] = useState(false);
+  const [showFullValue, setShowFullValue] = useState(false);
+  const [showFullUnits, setShowFullUnits] = useState(false);
```
tealefristoe, line 32: These seem like they're used for styling, and I wonder if it would be possible to switch up the css a bit so we could drop them. They modify the `.node` and `.value-unit-row`.

For `.node`, they adjust height--would it be possible to make `.node` always use `height: auto;`?

`.value-unit-row` also adjusts height--could this be switched to always use `auto`? `align-items` is also modified, but it's never set for the un-expanded version--could it always be set to `flex-start`?

src/diagram/components/ui/expandable-input.tsx
```diff
@@ -0,0 +1,108 @@
...
+  maxLength?: number;
+  placeholder?: string;
+  title: string;
+  value?: string | number;
+  handleBlur: () => void;
+  handleChange?: (evt: ChangeEvent<HTMLTextAreaElement>) => void;
+  handleFocus: (evt: FocusEvent<HTMLInputElement|HTMLTextAreaElement>|MouseEvent<HTMLInputElement|HTMLTextAreaElement>) => void;
+  handleKeyDown?: (evt: React.KeyboardEvent) => void;
+  handleToggle: () => void;
+  setRealValue: (value: number | undefined) => void;
```
tealefristoe, line 22: Since text inputs won't want this, you could make it optional and only render a `NumberInput` when `inputType === "number" && setRealValue`. In other words, if someone tries to render an `ExpandableInput` with `inputType === "number"` but without passing in a `setRealValue`, you can just render a `textarea` instead. If you go this route, I think it would make sense to add a comment explaining that `setRealValue` is required to render an expandable `NumberInput`.

src/diagram/components/quantity-node.tsx
```diff
@@ -201,26 +220,22 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
...
-              data-testid={"variable-expression"}
-              maxLength={kMaxNotesCharacters}
-              onChange={onExpressionChange}
-              placeholder={"expression"}
+          <div className={classNames("variable-info-row", "expression-row", {"expanded": (hasLongExpression && showFullExpression) })} data-testid="variable-expression-row">
+            <ExpandableInput
+              error={!!(variable.computedValueError || variable.computedUnitError)}
+              inputType="text"
+              lengthToExpand={kExpressionExpandLength}
+              maxLength={kMaxNameCharacters}
```
tealefristoe, line 228: Not sure what this should be but it needs to be longer. Maybe if `maxLength` is undefined there is no limit?

tealefristoe (commented):
I think this is a good direction, but it's still going to take a bit of work. I left some comments, but here are some high level thoughts:
- I think using `updateShowFullValue` for the `NumberInput` is good, but I suggested changing the name to something more generic because I think it could be used in other ways by other components.
- I discuss this in comments, but I'm hoping it'll be possible to either eliminate the `showFullX` and `hasLongX` state from `QuantityNode`, or at least move that state into `ExpandableInput` if it can't be completely removed. Note that I didn't tag this state for the expression in my comments, but I'm hoping that can follow the same pattern as value and units. Removing this state would greatly simplify `QuantityNode`.
- I found a bug with selection. If you select a field, then deselect the field by clicking elsewhere on the same card, the card becomes deselected and cannot be selected again until you click on a field on that card or something else in the diagram.
  - It's probably the same bug, but something similar happens with two cards. Select a field in card A, then click on card B. Neither card is selected, and you can't select card B until you either click one of its fields or elsewhere in the diagram.

emcelroy:
@tealefristoe Great suggestions -- thank you! I believe I followed them all in the latest commit and fixed that bug you found. 

I had actually started out trying `height: auto` for the cards and rows, but couldn't seem to make it work for some reason. I revisited the idea at your suggestion, though, and it seems to work perfectly fine. And like you said, that made some variables and classes unnecessary.

tealefristoe (approved):
This looks great! Thanks for taking the time to revise the code, I think it's a lot more streamlined now.

PR #43, opened 2022-12-02
Color Variable Chips to Match Card

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183915153

[#183915153]

These changes make variable chips' colors match the chips' associated variable cards. The colors should be applied in the text and drawing tiles, as well as the insert dialogs. There are also some other minor styling tweaks to match specs.

### DIFF
--- a/src/diagram/components/ui/variable-chip.scss
+++ b/src/diagram/components/ui/variable-chip.scss
@@ -1,10 +1,23 @@
-@use "../../../scss/vars.scss";
+@import "../../../scss/vars.scss";
+@import "../../../scss/mixins.scss";
 
 .variable-chip {
-  border: 1px solid black;
-  border-radius: 5px;
-  padding: 1px 3px;
-  margin: 0 1px;
+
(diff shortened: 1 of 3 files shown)


tealefristoe (approved):
Easy peasy! Looks great :D

src/diagram/components/ui/variable-chip.scss
```diff
@@ -1,10 +1,23 @@
-@use "../../../scss/vars.scss";
+@import "../../../scss/vars.scss";
+@import "../../../scss/mixins.scss";
```
kswenson, line 2: FYI: `@import` has been deprecated in favor of `@use`, so we should be migrating the other way moving forward. (I only learned about `@use` after seeing it in a previous PR of @tealefristoe's.)

emcelroy: @kswenson Good to know. Will keep that in mind going forward 👍

PR #45, opened 2022-12-02
Fix crash when variable card deleted via keyboard DELETE key

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183929844

[#183929844]

This change fixes an issue where deleting a source variable card by selecting it and pressing the DELETE key on your keyboard would cause a white screen crash.

### DIFF
--- a/src/diagram/components/diagram.tsx
+++ b/src/diagram/components/diagram.tsx
@@ -86,14 +86,14 @@ export const _Diagram = ({ dqRoot, getDiagramExport, hideControls, hideNavigator
   };
 
   const onElementsRemove = (elementsToRemove: Elements) => {
-    for(const element of elementsToRemove) {
+    for (const element of elementsToRemove) {
       if (isEdge((element as any))) {
         const edge = element as Edge;
         const { source, target } = edge;
         const targetModel = dqRoot.getNodeFromVariableId(target);
         const sourceModel = dqRoot.getNodeFromVariableId(source);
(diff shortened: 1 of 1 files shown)


kswenson (approved):
👍 This change seems uncontroversial. Do you think we're actually handling deletion properly at this point or is there more work to be done here?

tealefristoe (approved):
Looks good

kswenson (approved):

emcelroy:
@kswenson I think deletion is being handled properly. Elsewhere, Teale noted that ideally we'd be using the same function to delete cards whether they're deleted via the keyboard or the delete button in the UI (which wasn't affected by this bug). I agree, but we're not sure it's worth spending the time on that now. I was going to see what @scytacki thought, but would be glad to hear your thoughts as well. I lean toward just getting this change in to fix the bug, and then revisiting the idea of consolidating the delete functions later. But it'd be good to get other opinions.

kswenson:
> @kswenson I think deletion is being handled properly. Elsewhere, Teale noted that ideally we'd be using the same function to delete cards whether they're deleted via the keyboard or the delete button in the UI (which wasn't affected by this bug). I agree, but we're not sure it's worth spending the time on that now. I was going to see what @scytacki thought, but would be glad to hear your thoughts as well. I lean toward just getting this change in to fix the bug, and then revisiting the idea of consolidating the delete functions later. But it'd be good to get other opinions.

Fixing the immediate crasher is clearly the priority.

PR #46, opened 2022-12-05
Format static value and unit fields

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183570351

[#183570351] 

[Previous work on this story](https://github.com/concord-consortium/quantity-playground/pull/42) left out required style changes to the static value and unit "fields" displayed below an expression field. The static value and unit weren't actually inputs but DIVs made to look like inputs.

These changes use disabled versions of the value and unit input fields to display the static values. Doing so allows re-use of the expand/collapse feature/styling recently added for the input fields.

### DIFF
--- a/src/diagram/components/quantity-node.tsx
+++ b/src/diagram/components/quantity-node.tsx
@@ -185,7 +185,7 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
             />
           </div>
         }
-        {hasExpression ? renderValueUnitInput({disabled: true}) : renderValueUnitInput({disabled: false})}
+        {renderValueUnitInput({disabled: hasExpression})}
         <div className={classNames("variable-info-row", "description-row", { expanded: showDescription })}>
           {showDescription && 
             <TextareaAutosize
--- a/src/diagram/components/quantity-node.tsx
+++ b/src/diagram/components/quantity-node.tsx
@@ -141,19 +141,6 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
     );
   };
 
-  // const renderValueUnitUnEditable = () => {
-  //   return (
-  //     <div className="variable-info-row value-unit-row">
-  //       <div className={classNames("variable-info value static", {"no-value": !shownValue, "invalid": variable.computedValueError})}>
-  //         {shownValue !== undefined ? variable.computedValueWithSignificantDigits : "value"}
-  //       </div>
-  //       <div className={classNames("variable-info unit static", {"no-value": shownUnit, "invalid": variable.computedUnitError})}>
-  //         {shownUnit || "unit"}
-  //       </div>
-  //     </div>
-  //   );
-  // };
-
   const nodeHeight = hasExpression ? "155px" : "120px";
   const nodeWidth = "220px";
   const targetNodeHandleStyle = {height: nodeHeight, width: nodeWidth, left: "1px", opacity: 0, borderRadius: 0};
--- a/src/diagram/components/quantity-node.tsx
+++ b/src/diagram/components/quantity-node.tsx
@@ -103,22 +103,29 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
     data.dqRoot.setSelectedNode(data.dqRoot.selectedNode);
   };
 
-  const renderValueUnitInput = () => {
+  const renderValueUnitInput = (params: {disabled: boolean}) => {
+    const { disabled } = params;
+    const staticValue = shownValue !== undefined
+                          ? variable.computedValueWithSignificantDigits
+                          : undefined;
+
     return (
       <div className="variable-info-row value-unit-row">
         <ExpandableInput
+          disabled={disabled}
           error={!!(variable.computedValueError)}
           inputType="number"
           lengthToExpand={kDefaultExpandLength}
           maxLength={kMaxNameCharacters}
           placeholder="value"
           title="value"
-          value={variable.value}
+          value={!disabled ? variable.value : staticValue}
           handleBlur={handleFieldBlur}
           handleFocus={handleFieldFocus}
           setRealValue={variable.setValue}
         />
         <ExpandableInput
+          disabled={disabled}
           error={!!(variable.computedUnitError)}
           inputType="text"
           lengthToExpand={kDefaultExpandLength}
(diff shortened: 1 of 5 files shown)


src/diagram/components/quantity-node.tsx
```diff
@@ -191,7 +185,7 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
             />
           </div>
         }
-        {hasExpression ? renderValueUnitUnEditable() : renderValueUnitInput()}
+        {hasExpression ? renderValueUnitInput({disabled: true}) : renderValueUnitInput({disabled: false})}
```
tealefristoe, line 188:
```suggestion
        { renderValueUnitInput({disabled: hasExpression}) }
```

src/diagram/components/ui/expandable-input.tsx
```diff
@@ -70,6 +78,7 @@ export const ExpandableInput = ({
           autoComplete="off"
           className={inputClasses}
           data-testid={`variable-${title}`}
+          disabled={disabled}
```
tealefristoe, line 81: I don't think you need to pass disabled here because you already check to make sure disabled is false before you try to render a `NumberInput`.

tealefristoe (approved):
Looks good! I think it could go as is, but I made a couple of small suggestions, plus asked a question that I think is worth thinking about a bit. I could be wrong, but I think `hasLongValue` could be changed to a constant in `ExpandableInput`, which would simplify it a bit and probably make it a little more efficient.

src/diagram/components/quantity-node.tsx
emcelroy: Ah, nice 👍

src/diagram/components/ui/expandable-input.tsx
emcelroy: I'm pretty sure it's needed because the check for `!disabled` is part of a trio of conditions that have to be met to render a number input. Even if the input type is a string, it may be disabled or not disabled.

src/diagram/components/ui/expandable-input.tsx
```diff
@@ -84,16 +93,17 @@ export const ExpandableInput = ({
     }
   };
 
+  useEffect(() => {
+    if (disabled && value) {
+      setHasLongValue(!!(value.toString().length >= lengthToExpand));
```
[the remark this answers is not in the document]
emcelroy: Using a function seems like a solid suggestion, thanks, but for some reason I can't figure out, it doesn't work. If I try to check the value via a function, the fields won't expand when the value gets too long. I tried several variations on the function idea, but had no luck.

I believe `hasLongValue` has to be a state variable because otherwise it couldn't be set by the `useEffect` that's needed to update disabled fields. I'm not sure if that `useEffect` is the best option for making sure disabled fields are updated, but it's the only way I was able to make it work.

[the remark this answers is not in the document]
tealefristoe: I'd have to take a look at what you tried to know why the function didn't work. I think it should be possible to make it work by adding `const checkLongValue = (value: number | undefined) => value?.toString().length >= lengthToExpand;` near the top of the functional component (above the `useState`s), then use that function when you were directly checking the length (lines 29, 37, and 98).

I guess I'm not sure what the `useEffect` is doing. From my understanding, it will change `hasLongValue` when `disabled` or `value` changes (I don't think `lengthToExpand` should ever change). But those are props, so the component should re-render when either of them changes anyway. So it seems to me like you should be able to replace the `useState` that defines `hasLongValue` with a constant: `const hasLongValue = value?.toString().length >= lengthToExpand;`. However, it's definitely possible that I'm misunderstanding something.

(I thought that `!!` is necessary there, but it looks like it isn't because `undefined >= lengthToExpand` evaluates to false, not undefined. Good to know!)

tealefristoe (approved):
I added a couple of comments that I hope will clarify my previous comments. I'm happy to discuss more tomorrow if they're still confusing and you're interested.

That said, I think this all looks ok so you should feel free to move forward with it as is.

src/diagram/components/ui/expandable-input.tsx
emcelroy: Oh, I think maybe you need to expand the code in this view to see the entire `if...else`? It doesn't pass `disabled` to `NumberInput`, only to the `textarea` in the `else` part.

emcelroy:
@tealefristoe Thanks for the additional comments. I'm not sure what I did differently this time, but the length check is now working in a function. I thought I had tried exactly what you described before, plus multiple variations and none of them worked. This morning it seems to work fine, though 🤷🏻‍♂️

I'm still not able to make things work without making `hasLongValue` a state variable. I added a comment clarifying why the `useEffect` is necessary. Does that make at least a little more sense?

tealefristoe (approved):
You're totally right about that `disabled` prop. Sorry about that.

I still feel like `hasLongValue` should work as a const, but it's not a big deal (plus I could be wrong!), so I think you should merge as is.

Thanks for the back and forth on this review, I learned a few things!

src/diagram/components/ui/expandable-input.tsx
```diff
@@ -84,16 +98,20 @@ export const ExpandableInput = ({
...
 
+  // Ensure that hasLongValue is updated when the value changes in 
+  // disabled fields -- the onChange handler above does not get triggered
+  // by disabled fields.
+  useEffect(() => {
+    if (disabled && value) {
+      setHasLongValue(isLongValue(value, lengthToExpand));
+    }
+  }, [disabled, lengthToExpand, value]);
+
```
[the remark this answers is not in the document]
tealefristoe: Good insight about the `useEffect`, that makes sense why it would be necessary.

Your question about `onChange` vs `onBlur` in `NumberInput` is a great one, and unfortunately I don't have a great answer. The main reason I remember using `onBlur` rather than `onChange` for the `NumberInput` is that's when PIs requested the value change. I can think of some possible technical reasons to not to it, but I'm not sure how legitimate those reasons are. So I think it would be a good idea to experiment with switching when `NumberInput` updates its value. If you do, you should test the change in the new/edit variable dialogs as well as the variable card.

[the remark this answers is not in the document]
kswenson: This is basically the distinction between controlled (`onChange`) vs. uncontrolled (`onBlur`) components. If you don't need external access to the value between focus and blur, then uncontrolled is generally simpler. If you _do_ need external access to the value, then a controlled component is generally the solution. You're essentially creating a semi-controlled component here by not putting the value in state but then putting `hasLongValue` in state. I suspect you could eliminate `hasLongValue` from state by having the value in state instead, thus switching to a controlled component. Not sure whether it's worth it, though, so I leave that up to you. 😉

PR #47, opened 2022-12-06
Refactor method to prevent inadvertent dragging.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183644075

[#183644075]

Previously, we used `event.preventDefault()` when an input field was being brought into focus in order to prevent inadvertent dragging of the field's parent node when the user tried to select the field's text. Then we would use `setSelectedNode()` to set the field's parent node to be selected since `event.preventDefault()` prevented that from happening. This caused an issue where the selected node could not always reliably be unset when the field was blurred.

This approach uses React Flow's `draggable` node property, and doesn't require `event.preventDefault()`. So the selected state of nodes is properly set and unset.

### DIFF
--- a/src/diagram/components/quantity-node.tsx
+++ b/src/diagram/components/quantity-node.tsx
@@ -89,18 +89,15 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
   };
 
   const handleFieldFocus = (evt: FocusEvent<HTMLInputElement|HTMLTextAreaElement>|MouseEvent<HTMLInputElement|HTMLTextAreaElement>) => {
-    // Stopping propagation allows the user to select text in the input field without 
-    // inadvertently dragging the card/node.
-    evt.stopPropagation();
-    data.dqRoot.setSelectedNode(data.node);
+    // Setting draggable to false allows the user to select text in the input field
+    // without inadvertently dragging the card/node.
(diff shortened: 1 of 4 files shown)


scytacki (changes requested):
A problem with this approach is that the user can't drag when they have a field focused. The mouse cursor changes to a hand which makes it seem like they could drag. They have to un focus first before they can drag.

Perhaps that is a trade off we want, just to fix this. If we do that then hopefully we can figure out how remove the hand cursor when the node isn't draggable. 

But ideally it would be better to let the user drag the node while the field is focused. I guess that would mean going back to the stopPropagation approach.

emcelroy:
@scytacki Thanks for catching that. Good point.

I think the solution is undo the changes I added in a previous PR to set the parent node to selected when one of its child fields is in focus. I now see that's unnecessary and based on a mis-interpretation of the specs. I'll make a new branch and PR for that.

tealefristoe (changes requested):
I agree with Scott's concerns, but I think this is an improvement over what we were doing before. How difficult would it be to only show the grabber hand when the node is actually draggable? Seems like it would be pretty easy to add one more class based on the draggable status and only change the cursor if the draggable class is present.

UPDATE: I now see we're moving a different direction for this, which looks good to me.

PR #50, opened 2022-12-12
Fix crash when variable deleted by keyboard DELETE

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183929844

[#183929844]

Similar to [this fix for deleting a source card](https://github.com/concord-consortium/quantity-playground/pull/45), this will fix an issue where deleting a target/expression card by selecting it and pressing the DELETE key on your keyboard would cause a white screen crash.


tealefristoe (approved):

kswenson (approved):
👍 LGTM

PR #53, opened 2023-01-20
feat: add spellcheck=false to fields in edit variable dialog (PT-183914399)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183914399

These changes disable spellcheck and autocomplete (aka, autofill, autosuggest) for fields in the New/Edit Variable dialog. (An [earlier PR](https://github.com/concord-consortium/quantity-playground/tree/183914399-disable-spell-check) did the same for the fields in the variable cards.) Note that spellcheck should be enabled for the Notes/description field.

### DIFF
--- a/src/diagram/components/dialog/edit-variable-dialog.tsx
+++ b/src/diagram/components/dialog/edit-variable-dialog.tsx
@@ -36,6 +36,7 @@ export const EditVariableDialogContent = observer(({ variable }: IEditVariableDi
         maxCharacters={kMaxNotesCharacters}
         rows={2}
         setValue={variable.setDescription}
+        spellCheck={true}
         value={variable.description || ""}
       />
       <TextRow inputId="evd-units" label="Units" value={variable.unit || ""} setValue={variable.setUnit} width={230} />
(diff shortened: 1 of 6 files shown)


src/diagram/components/dialog/edit-variable-dialog.tsx
```diff
@@ -36,6 +36,7 @@ export const EditVariableDialogContent = observer(({ variable }: IEditVariableDi
         maxCharacters={kMaxNotesCharacters}
         rows={2}
         setValue={variable.setDescription}
+        spellCheck={true}
```
emcelroy, line 39: This is for the Notes fields for which we do want to allow spellcheck

src/diagram/components/quantity-node.tsx
```diff
@@ -176,6 +176,7 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
         <div className={classNames("variable-info-row", "description-row", { expanded: showDescription })}>
           {showDescription && 
             <TextareaAutosize
+              autoComplete="off"
```
emcelroy, line 179: Another instance of the Notes fields for which we do want to allow spellcheck. So we only disable autocomplete here.

PR #54, opened 2023-01-31
feat: Different Arrow for unused variables (PT-184037417)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184037417

Adds multiple styles for connecting lines (aka edges) between nodes. The main purpose is to differentiate inputs used in a target node's expression from those that aren't used in the expression. The two kinds of edges are different in color and line type. See PT story and screenshot below for details.

Three custom SVG marker ends are now defined in the `Diagram` component which are then utilized in diagram.scss. The marker ends are essentially the same except in color.

There is also a new `ConnectionLine` component used to render edges when they are first created, before they are connected to a target node. Unconnected edges are differentiated from connected edges by their color (blue).

These changes could very possibly be simplified after we update to a new version of React Flow. (I attempted an update but ran into a problem that I couldn't solve in the allotted time.)

image

### DIFF
--- a/src/diagram/components/floating-edge.tsx
+++ b/src/diagram/components/floating-edge.tsx
@@ -1,7 +1,6 @@
 import React, { useMemo } from "react";
 import { getBezierPath, useStoreState } from "react-flow-renderer/nocss";
 import { getEdgeParams } from "../../utils/diagram/floating-edge-util";
-import classNames from "classnames";
 
 interface IProps {
   id: string;
@@ -13,8 +12,6 @@ export const FloatingEdge: React.FC<IProps> = ({ id, source, target }) =>  {
   const nodes = useStoreState((state) => state.nodes);
   const sourceNode = useMemo(() => nodes.find((n) => n.id === source), [source, nodes]);
   const targetNode = useMemo(() => nodes.find((n) => n.id === target), [target, nodes]);
-  const sourceVariableName = sourceNode?.data.node.variable?.name;
-  const isUsedInExpression = targetNode?.data.node.variable?.expression?.includes(sourceVariableName) || false;
 
   if (!source || !target) {
     return null;
@@ -32,13 +29,9 @@ export const FloatingEdge: React.FC<IProps> = ({ id, source, target }) =>  {
     targetY,
   });
 
-  const edgeClasses = classNames(
-    "react-flow__connection",
-    { "source-used-in-target-expression": isUsedInExpression }
-  );
   // used the react-flow__edgeupdater class because it has some react-flow-renderer event handler that allows the edge to be deleted
   return (
-    <g className={edgeClasses}>
+    <g className="react-flow__connection">
       <path id={id} className="react-flow__edge-path react-flow__edgeupdater" d={d}/>
     </g>
   );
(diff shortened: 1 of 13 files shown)


src/diagram/models/mathjs-utils.ts
```diff
@@ -29,19 +29,8 @@ export const getMathUnit = (value: number, unitString: string): math.Unit | unde
...
 export const replaceInputNames = (expression: string, inputNames: (string | undefined)[]) => {
-  let localExpression = expression;
-
-  inputNames.forEach((name, index) => {
-    if (!name) {
-      return;
-    }
-    // A RegExp is used so all matches are replaced not just the first one
-    localExpression = localExpression.replace(RegExp("`" + name + "`", "g"), `input_${index}`);
-  });
```
emcelroy, line 44: This was meant to support variable names with spaces and the use of backticks in equations to utilize such variable names. We're going to remove it since we no longer support variables with spaces in the names. Any old equations that use such variables will throw an error, but the user can just edit the variable names and remove the backticks to clear that it.

scytacki (changes requested):
I'd suggest changing the logic so the set of usedInputs is computed by the Variable model. This keeps the separation of the variable logic from the diagram representation of it. It also means that other views of the Variable model can show which inputs are used. We don't have any other views of the Variable model that need to now this now, but I could see it being requested in the future.

src/diagram/components/floating-edge.tsx
```diff
@@ -1,33 +1,45 @@
...
   target: string;
 }
-export const FloatingEdge: React.FC<IProps>  = ({ id, source, target }) =>  {
+
+export const FloatingEdge: React.FC<IProps> = ({ id, source, target }) =>  {
   const nodes = useStoreState((state) => state.nodes);
   const sourceNode = useMemo(() => nodes.find((n) => n.id === source), [source, nodes]);
   const targetNode = useMemo(() => nodes.find((n) => n.id === target), [target, nodes]);
+  const sourceVariableName = sourceNode?.data.node.variable?.name;
+  const isUsedInExpression = targetNode?.data.node.variable?.expression?.includes(sourceVariableName) || false;
```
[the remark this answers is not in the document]
emcelroy: @scytacki Thanks. That all makes a lot of sense.

One thing I found that made it a little difficult was apparently not all of an edges' properties are passed down to the FloatingEdge component. Many are, but some aren't for no reason that's apparent to me. Kind of critically, `className` is one of the properties that isn't passed. `style` is passed, but I didn't think it'd be good to define a bunch of style rules in the variable model.

Luckily, the `className` value for an edge does get added to a parent of the related FloatingEdge elements in the HTML. So I was able to get the desired result by modifying the stylesheet rules accordingly.

src/diagram/models/mathjs-utils.ts
```diff
@@ -50,12 +40,21 @@ export const replaceInputNames = (expression: string, inputNames: (string | unde
...
   }
 };
+
+export const replaceInputNames = (expression: string, inputNames: (string | undefined)[]) => {
+  return parseExpression(expression, inputNames).expression;
+};
+
+export const getUsedInputs = (expression: string, inputNames: (string | undefined)[]) => {
+  return parseExpression(expression, inputNames).inputsInExpression;
+};
```
emcelroy, line 60: Does this seem pretty straightforward? I didn't want to duplicate a lot of the original `replaceInputNames` function's code, but I'm not sure this is the clearest way to do that.

scytacki: Yes this looks good.

scytacki (approved):
Yes this looks good to me. 👍

PR #56, opened 2023-02-06
fix: Different arrow for unused variables. (PT-184037417)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184037417

[Recent changes](https://github.com/concord-consortium/quantity-playground/pull/54) to style input arrows based on whether they're used in an expression could result in a used input's arrow being styled as if it's _not_ used if an invalid expression like `a+b+` is entered. The current code won't build a list of used inputs when Math.js's `parse` function returns an error. These changes would build that list using an alternate method when `parse` fails.


scytacki (approved):
Looks good to me. 👍

PR #58, opened 2023-02-08
feat: Edit Variable dialog displays read-only expressions (PT-183974159)

emcelroy:

With these changes, the variable edit dialog form will display a read-only Expression field. It also adds a read-only calculation field, makes the Unit and Value fields read-only for expression variables, and implements some new styling for the dialog form.

Once these changes are finalized and merged, [these related changes to CLUE](https://github.com/concord-consortium/collaborative-learning/pull/1566) will be required.


src/diagram/components/dialog/edit-variable-dialog.tsx
```diff
@@ -1,74 +1,145 @@
...
   const variableClone = useMemo(() => Variable.create(getSnapshot(variable)), [variable]);
+  const hasInputs = variable.numberOfInputs > 0;
+  const hasExpression = !!variableClone.expression;
+
+  // Because variableClone was created from a snapshot, we will not be able to access 
+  // computed values related to the original variable's inputs. We need to get those values
+  // from the original variable and either update variableClone or pass the values to the
+  // EditVariableDialogContent component separately.
+  const calculationString = variable.calculationString;
+  const errorMessage = variable?.computedValueError || variable?.computedUnitError || variable?.computedUnitMessage;
```
emcelroy, line 120: I'm not sure about how I handled this. I'm specifically wondering if it would be better to add some actions to the Variable model so we can set the error message and calculation string that way (like I use the existing `setValue` and `setUnit` below) instead of passing down values through new props I added to the EditVariableDialogContent component.

src/diagram/components/dialog/edit-variable-dialog.tsx
```diff
@@ -1,74 +1,127 @@
...
+export interface IEditVariableDialogContent {
+  variable?: VariableType;
+  variableClone: VariableType;
 }
-export const EditVariableDialogContent = observer(({ variable }: IEditVariableDialogContent) => {
+
+export const EditVariableDialogContent = observer(({ variable, variableClone }: IEditVariableDialogContent) => {
+  // Because variableClone was created from a snapshot, we will not be able to access 
+  // computed values related to the original variable's inputs. We need to get those values
+  // from the original variable.
```
tealefristoe, line 24: I think it would be worth explaining why we have a variable clone in the first place: so the user can modify its properties directly, but those changes won't be saved unless the ok button is pushed.

src/diagram/components/dialog/edit-variable-dialog.tsx
```diff
@@ -1,74 +1,127 @@
...
+  const errorMessage = variable?.computedValueError || variable?.computedUnitError || variable?.computedUnitMessage;
+  const isExpressionVariable = variable?.hasInputs;
+
   const updateName = (newName: string) => {
-    variable.setName(processName(newName));
+    variableClone.setName(processName(newName));
+  };
+  const updateValue = (newValue: number | undefined) => {
+    variableClone.setValue(newValue);
   };
```
tealefristoe, line 33: Would it be possible to directly pass `variableClone.setValue` as the `setRealValue` prop below, instead of wrapping it in `updateValue`? I know this was to fix an issue with testing, but I think it's worth another shot, because I don't think the wrapper is actually doing anything.

tealefristoe (changes requested):
Looks good! I just made a couple of comments that should be quick to address.

Also, I thought it was worth pointing out that I'm the guy who adds that extra newline in the imports 😃 There's a method to my madness, though. The newline separates external library imports from our own files. But you don't seem to be a fan of those newlines, and I'm ok scrapping them.

src/diagram/components/dialog/number-row.tsx
```diff
@@ -1,28 +1,35 @@
 import React from "react";
-
 import { DialogRow } from "./dialog-row";
 import { NumberInput } from "../ui/number-input";
 import { isValidNumber } from "../../utils/validate";
+import classNames from "classnames";
```
kswenson, line 5: Move import ⬆️ to keep external imports together

kswenson (commented):
I don't feel strongly about whether there should be a newline between the external imports and the internal ones, but I do think they should be grouped.

src/diagram/components/dialog/edit-variable-dialog.tsx
emcelroy: Good point. I've updated the comment.

src/diagram/components/dialog/edit-variable-dialog.tsx
emcelroy: Yes, thanks for pointing that out. I've made that change.

src/diagram/components/dialog/number-row.tsx
emcelroy: @kswenson @tealefristoe Thanks. I didn't see the pattern, but get it now.

I've left in or added the separating newline in the latest commit. I'm kind of ambivalent about the new line, too, but I don't mind it now that I know the purpose. We have a somewhat similar pattern with separating the stylesheet imports with a newline. Seems fine to me.

tealefristoe (approved):
Looks great!

scytacki (approved):
Since this looks good to the rest of you, I'm going to approve too. I didn't look closely at it.

PR #60, opened 2023-02-14
fix: Edit Variable dialog displays read only expressions (PT-183974159)

emcelroy:

This fixes two issues related to [this previous PR](https://github.com/concord-consortium/quantity-playground/pull/58).

1) The Name field was not expanding to the full width of its container.

2) The Unit field was not being limited to 27 characters.

### DIFF
--- a/src/diagram/components/dialog/dialog.scss
+++ b/src/diagram/components/dialog/dialog.scss
@@ -77,6 +77,7 @@ $invalid-outline: .5px solid vars.$invalid-red;
   max-width: calc(100% - 15px);
   overflow-y: scroll;
   padding: 3.5px 6px;
+  width: 100%;
 
   &.invalid {
     border: 1.5px solid vars.$invalid-red;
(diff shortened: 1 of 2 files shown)


tealefristoe (approved):
Looks good!

I wanted to check this out in the browser, but I wasn't sure how to reach an in development branch for quantity-playground. I don't have major concerns about this simple PR, but I did want to check the effect of `width: 100%` for `.dialog-input`.

emcelroy:
> I wanted to check this out in the browser, but I wasn't sure how to reach an in development branch for quantity-playground. I don't have major concerns about this simple PR, but I did want to check the effect of `width: 100%` for `.dialog-input`.

@tealefristoe I think the only way to do that right now would be to check out the branch in a local clone of the repository, then do `npm install` (if needed) and then `npm start`.

FWIW, here's what the field looks like with this change. Previously, the Name field was only taking up about 80-85% percent of the column width.

image

PR #61, opened 2023-02-15
184258400 More Missing Operands

tealefristoe:
This PR adds more cases for the missing operand warning message, as well as new cypress tests for them.

### DIFF
--- a/cypress/integration/error-messages.test.ts
+++ b/cypress/integration/error-messages.test.ts
@@ -31,15 +31,15 @@ context("Test error messages", () => {
       errorBottom().should("contain.html", incompleteExpanded);
     });
 
-    it("renders incomplete for '+ a + b'", () => {
-      enterExpression("+ a + b");
-      computedNode().find(".error-short").should("contain.html", "Um, still working?");
-    });
-
-    it("renders incomplete for '+a'", () => {
-      enterExpression("+a");
-      computedNode().find(".error-short").should("contain.html", "Um, still working?");
-    });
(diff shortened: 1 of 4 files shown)


tealefristoe:
I'm not sure why this is complaining about test coverage--I explicitly added tests to cover all of the new cases. Maybe the problem is that the tests have been added in a loop instead of hard coded directly?

emcelroy (approved):
Looks good to me! I think you're right that Codecov isn't "smart" enough to count your new tests toward the coverage percentage. If you're sure they're being run and passing, I think it's fine to ignore that.

cypress/integration/error-messages.test.ts
```diff
@@ -31,15 +31,15 @@ context("Test error messages", () => {
...
-    });
+    const testIncomplete = (expression: string) => {
+      it (`renders incomplete for "${expression}"`, () => {
+        enterExpression(expression);
+        computedNode().find(".error-short").should("contain.html", "Um, still working?");
+      });
+    };
+    const incompleteStrings = ["+ a + b", "+a", "*a+b", "a+*b", "/a+b", "a+/b", "^a+b",
+      "a+^b", "(a+b", "a(+b", "a+(b", "a+b)"];
+    incompleteStrings.forEach(expression => testIncomplete(expression));
```
kswenson, line 42: I think the `it()` needs to be at the top level and the looping should happen within it.

tealefristoe: Thanks, that fixed the issue.

PR #62, opened 2023-02-18
183249829 Unit Conversion Issues

tealefristoe:
This PR attempts to tackle multiple issues with units and conversion between them.
- Remove many built-in units.
- Improve infrastructure for adding custom units, including the ability to add built-in units.
- Give each variable its own math library, which refresh whenever custom units are added.

Additional hodgepodge upgrades:
- Stop considering unit messages errors, to avoid the unit canceling message.
- Stop shifting expandable input content to left.


scytacki (commented):
If this works it seems fine to me.

I didn't look super close at the code, but I would think there might be a problem with different math instances: the input units will not be added to the math instance that is computing the expression.  However perhaps this works because I think the expression works on the `Unit` object from the input card. In this case the `Unit` object is a combination of the value and unit. So perhaps passing a `Unit` object from one math instance to another automatically adds the unit of the `Unit`. 
One test of this might be to set the input card of `a` to be `2 cats` and then have the output card be `a + 3 cats`.  If that works then you could try the more complicated `a + 1 cat`. 

If this is a problem, I still think the separate math instances could work. We'd just have to keep a global list of custom units and make sure they get added to all of the math instances before the expression is evaluated. And perhaps this also means the computed value of an output card should depend on this global unit list, that way the value will be recomputed if the  unit is used anywhere else.

emcelroy (approved):
Looks good 👍

scytacki (approved):
looks good to me

Maybe I missed them but it would be nice to have some tests that require the new customUnitArray support. No need to hold up this PR though.

scytacki:
looks like you were adding those tests just as I was reviewing it. :)

PR #63, opened 2023-02-21
fix: Edit Variable dialog displays read only expressions (PT-183974159)

emcelroy:

These changes specifically prevent line breaks in the variable edit dialog's Value and Unit fields whch was a follow up request made in a comment on the above PT story. Also included is a small style change to prevent scrollbars from appearing in textareas that do not yet have scrollable content.


tealefristoe (approved):
Looks good!

I tried to think of a way to combine the duplicated code between `TextAreaRow` and `NumberInput`, but I couldn't think of a good way to do it.

emcelroy:
> I tried to think of a way to combine the duplicated code between `TextAreaRow` and `NumberInput`, but I couldn't think of a good way to do it.

Yeah, I thought about that, too, but obviously didn't have any good ideas.

PR #64, opened 2023-02-24
Fix unit sluggishness

tealefristoe:
This PR fixes problems introduced with recent changes to how the mathjs library is replicated between different variables and updated. Those changes fixed a number of issues, but also made the diagram view run sluggishly, especially in the context of CLUE where many hidden variables are created.

- Only introduce new units when the unit field of a variable card is blurred, not on every keystroke.


emcelroy (approved):
Looks good!

PR #66, opened 2023-02-25
Fix some crashes around units

scytacki:
'$' was not being added a valid character before the units were added.
the overloaded createUnit wasn't working properly

PT-184564582


tealefristoe (approved):
Looks good! Thanks for tracking down that bug, cleaning up the typing, and adding some tests.

PR #67, opened 2023-02-27
fix: Clicking card fields should enable the edit button (PT-184565921)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184565921

Clicking into a variable card's fields wasn't causing the card to become selected which is the desired behavior. Among other features, when a card is selected, the Edit Variable button in the toolbar becomes enabled.

This change ensures that the card becomes selected when one of its fields comes into focus, causing the Edit Variable button to become enabled.


tealefristoe (approved):
Looks good 👍

PR #68, opened 2023-02-27
fix: Long values in variable dialog not displayed properly on card (PT-184474236)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184474236

This change makes the ExpandableInput component properly update its styling for long values after the values are set using the Edit Variable dialog. Previously, if a long value was set in the Edit Variable dialog, the input field was not changing to indicate that the value was longer than the field.


tealefristoe (approved):
Looks good!

PR #69, opened 2023-03-06
fix: Expression variable behaviour (PT-184616625)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184616625

BUG: When a variable card is selected and the Edit Variable dialog is opened but none of its fields are in focus, hitting the keyboard Delete key results in the selected variable card being deleted.

This change prevents the bug by disabling the default Delete key behavior for variable cards when a dialog is present.


tealefristoe (approved):
Looks good!

PR #70, opened 2023-03-21
fix: Expression Variable behaviour (PT-184616625)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184616625

### DIFF
--- a/patches/react-flow-renderer+9.7.4.patch
+++ b/patches/react-flow-renderer+9.7.4.patch
@@ -1,0 +1,29 @@
+diff --git a/node_modules/react-flow-renderer/dist/nocss/ReactFlow-nocss.esm.js b/node_modules/react-flow-renderer/dist/nocss/ReactFlow-nocss.esm.js
+index bc5be37..812ede3 100644
+--- a/node_modules/react-flow-renderer/dist/nocss/ReactFlow-nocss.esm.js
++++ b/node_modules/react-flow-renderer/dist/nocss/ReactFlow-nocss.esm.js
+@@ -2735,7 +2735,23 @@ var useKeyPress = (function (keyCode) {
+   useEffect(function () {
+     if (typeof keyCode !== 'undefined') {
+       var downHandler = function downHandler(event) {
+-        if (!isInputDOMNode(event) && (event.key === keyCode || event.keyCode === keyCode)) {
++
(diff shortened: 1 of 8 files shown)


patches/react-flow-renderer+9.7.4.patch
```diff
@@ -0,0 +1,29 @@
...
++        // selected node or edge before allowing it to affect the selected node or edge.
++        // That way if the diagram is a child element of a larger document, keydown
++        // events on elements outside the diagram will not cause unexpected behavior,
++        // such as deleting the currently selected node inside the diagram
++        // when you hit the Delete key while a modal window is open.
++        const activeNode = document.querySelector(".react-flow__node.selected");
++        const activeEdge = document.querySelector(".react-flow__edge.selected");
++        const isSelectedNodeOrEdge = (
++          (activeNode && activeNode.contains(event.target)) ||
++          (activeEdge && activeEdge.contains(event.target))
```
kswenson, line 23: should be equivalent to:
```suggestion
+          activeNode?.contains(event.target) || activeEdge?.contains(event.target)
```

kswenson (approved):
👍 LGTM

patches/react-flow-renderer+9.7.4.patch
emcelroy: @kswenson Thanks. I did consider that but wondered if it was [safe to use optional chaining](https://caniuse.com/mdn-javascript_operators_optional_chaining) in a .js file since support in browsers seems _relatively_ new. I didn't see any instances of it being used elsewhere in the .js files.

You think it's safe to use, though?

kswenson: That's a good point. I'm used to thinking in terms of transpiled code. That said, another approach occurred to me last night. What do you think of:
```suggestion
+          event.target.closest(".react-flow__node.selected") || event.target.closest(".react-flow__edge.selected")
```
and then you wouldn't need to the `querySelector` calls.

emcelroy: @kswenson Ah yeah, that's better 👍 Unfortunately, I already merged the existing fix in earlier today since the next CLUE release is fast approaching and I had to make the fix available for testing ASAP.

PR #71, opened 2023-04-18
184931580 Fix singular/plural unit bug

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/184931580

The project team discovered a crashing bug related to adding new units to mathjs. When adding a new unit, our software would check to make sure the unit was not already in the mathjs system, but it would not check the plural/singular version of the unit. So it was possible to attempt to add a singular/plural version that was already in the system, even though the original unit was not.

### DIFF
--- a/diagram-view/package.json
+++ b/diagram-view/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@concord-consortium/diagram-view",
-  "version": "0.0.41",
+  "version": "0.0.42",
   "description": "Concord Consortium Diagram View",
   "main": "dist/index.js",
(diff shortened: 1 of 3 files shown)


emcelroy (approved):
Looks good! I see Codecov complaining about lines 24 - 25 of src/diagram/utils/mathjs-utils.ts not being covered, but maybe it'd be tricky to test for a singular/plural unit that already exists?

tealefristoe:
@emcelroy Thanks, it wasn't difficult to add a test. I just needed a little nudge to do it :)

PR #72, opened 2023-04-19
fix: Undefined in expression calculation (PT-184955306)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184955306

Adds a check to `calculationString` to make sure each node has an associated input. If not, it simply returns the node instead of attempting to return a non-existent value and unit. This fixes a bug where, for example, `calculationString` would convert an expression like `a+b in ft` to `10 cm + 10 mm in undefined undefined`. It will now correctly convert it to `10 cm + 10 mm in ft`.


tealefristoe (approved):
👍

PR #73, opened 2023-05-24
184375687 Restrict Variable Names

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/184375687

This PR ensures that variable names can only include letters, digits, and underscores, and that they must begin with a letter. Characters entered into a variable card or dialog that do not meet these requirements are ignored.


scytacki (approved):
looks good to me

PR #74, opened 2023-05-27
185102033 Update Reactflow

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/185102033

This PR updates `react-flow-renderer` v9 to `reactflow` v11.


diagram-view/package.json
```diff
@@ -88,7 +88,6 @@
   },
   "dependencies": {
     "iframe-phone": "^1.3.1",
-    "mathjs": "^10.4.1",
-    "react-flow-renderer": "^9.7.4"
+    "mathjs": "^10.4.1"
```
emcelroy, line 91: Should `"reactflow": "^11.7.2"` be listed here.

src/diagram/components/diagram.tsx
```diff
@@ -1,25 +1,24 @@
...
 import { ConnectionLine } from "./connection-line";
 import { MarkerEnd } from "./marker-end";
 
-// We use the nocss version of RF so we can manually load
-// the CSS. This way we can override it.
-// Otherwise RF injects its CSS after our CSS, so we can't
-// override it.
-import "react-flow-renderer/dist/style.css";
-import "react-flow-renderer/dist/theme-default.css";
+import "reactflow/dist/style.css";
```
emcelroy, line 17: Sorry, I left this and the following CSS file import here, but re-reading the deleted comment, I'm thinking dropping these two lines may be possible now. Did you try leaving them out?

emcelroy: Forget I said the above. I tried removing the lines and doing so definitely breaks things.

It doesn't quite make sense to me that these lines would help if React Flow is already injecting the CSS, but apparently it does. I see they do something similar in [their examples of custom nodes](https://reactflow.dev/docs/guides/custom-nodes/), so looks like the right thing to do.

emcelroy (approved):
This looks good to me 👍 I had one question about something I'm not sure about in diagram-view/package.json. If you're sure that's how it should be, though, I approve.

src/diagram/components/quantity-node.tsx
```diff
@@ -243,6 +243,6 @@ const _QuantityNode: React.FC<IProps> = ({ data, isConnectable }) => {
 //
 // But if the model gets changed without a flow re-render
 // then, it doesn't update without the observer
-export const QuantityNode = observer(_QuantityNode);
+export const QuantityNode = memo(observer(_QuantityNode));
```
scytacki, line 246: Can you update the comment above about why this `memo` is now necessary?
Re-reading my comment, I think I meant to say "... but when I used it then the component wasn't updating..." key part was -> wasn't.

scytacki (changes requested):
This looks good.  I just had a question about the use of memo.

src/diagram/components/quantity-node.tsx
tealefristoe: This is actually a change that @emcelroy made before handing off the PR to me, so I'm not sure of why it was changed.

emcelroy: @scytacki @tealefristoe During the v9 to v10 migration, I followed the [Migrate to v10 guide](https://reactflow.dev/docs/guides/migrate-to-v10/). There's a section of it called [Memoize your custom nodeTypes and edgeTypes](https://reactflow.dev/docs/guides/migrate-to-v10/#2-memoize-your-custom-nodetypes-and-edgetypes). The main point is:

_Whenever you pass new node or edge types, we create wrapped node or edge component types in the background. This means that you should not create a new `nodeType` or `edgeType` object on every render. Memoize your nodeTypes and edgeTypes or define them outside of the component when they don't change._

Not doing so can lead to "bugs and performance issues."

I guess I was confused by the comment and probably intended to ask about it but forgot. FWIW, I didn't see any issues after adding the use of `memo` during the v10 upgrade. Everything seemed to work as expected. Unfortunately, I don't quite understand the nuances of how `observer` and `memo` may interact/conflict with each other. So I may be missing something there.

If you see no reason to not use `memo` and there really are no problems, then I would replace the comment on lines 233-237 with something like:

```
React Flow recommends memoizing custom node types to avoid possible bugs and performance issues.
```

Does that sound good? I can add a commit with that change, if so.

tealefristoe: Alright I just went ahead and updated the comment with the link you included above. I actually don't think memoizing is necessary, but it doesn't seem to hurt. I don't think it's necessary for two reasons:
- We aren't memoizing the edge types, and that doesn't seem to be causing problems.
- We're defining `nodeTypes` and `edgeTypes` outside of the component (see `diagram.tsx` lines 25-30), which is an alternate solution recommended in the react flow docs.

emcelroy: Oh, I see. I was misinterpreting their suggestion. Re-reading it and looking over the example from the migration guide, I see we are doing what they suggest in terms of defining the types outside the component. Thanks for pointing that out.

So you're right that memoizing isn't necessary. I think you could remove that and the comment.

kswenson: There is some confusion in this conversation. The docs recommend the use of `useMemo()` for components defined dynamically, which as previously stated is unnecessary for components defined at global scope. `memo` is a completely different concept, analogous to `PureComponent` back in the class component days, which indicates that a component should not be re-rendered if its props haven't changed. It's a performance optimization that allows children to avoid re-rendering when their parents re-render.

tealefristoe: Thanks for the clarification @kswenson. I went ahead and removed the call to `memo` altogether. I'm not sure what its impact was, positive or negative, but this way `QuantityNode` is more in line with our other components.

PR #76, opened 2023-05-31
Use concord version of mobx-state-tree.

tealefristoe:
Updates `quantity-playground` and `diagram-view` to use Concord Consortium's version of `mobx-state-tree`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
-        with:
-          node-version: '16'
       - name: Install Dependencies
         run: npm ci
       - name: Build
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -9,6 +9,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
```
kswenson, line 13: If locking the node version is required, a comment indicating what happened when it wasn't locked could be useful.

kswenson (approved):
👍 LGTM!

.github/workflows/ci.yml
tealefristoe: Thanks for pointing this out, it looks like it might not be necessary. I'm going to release a version without it.

PR #78, opened 2023-06-01
185298570 Make it easier to connect nodes

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/185298570

This PR makes it easier to establish a connection between two nodes by expanding the target area to drop a new arrow/connection/edge to be larger than the node. It also does a better job of approximating the height of a node.


tealefristoe:
The changes in this PR are being handled by another PR.

PR #79, opened 2023-06-01
185298547 Improve selecting arrows

tealefristoe:
PT stories:
https://www.pivotaltracker.com/story/show/185298547
https://www.pivotaltracker.com/story/show/185298503
https://www.pivotaltracker.com/story/show/185298532
https://www.pivotaltracker.com/story/show/185298570
https://www.pivotaltracker.com/story/show/184562862
https://www.pivotaltracker.com/story/show/184196685

(Sorry this PR is so large. It started as multiple smaller PRs, but I ended up combining them because they built off of each other. I didn't want to merge the smaller PRs into main until the `reactflow` update was released.)

This PR improves arrows/connections/edges between nodes in several ways.
- Arrows that are actively being created now look better because they use Bezier curves.
- The target for connecting nodes together now more accurately represents a node's size, making it easier to establish a connection.
- Illegal connections (connections to self, repeat connections, connections to ancestor) are now prevented.
- Illegal connections are visually communicated by highlighting the illegal target in red.
- There is a more substantial target for selecting existing arrows. This was particularly a problem for dashed arrows previously.
- The cursor in now a pointer and the arrow will change color when the user hovers over a selectable arrow.
- Selected arrows are now solid blue instead of a dark gray.
- A delete button has been added to hovered and selected arrows. Clicking this button when the arrow is selected will delete the arrow (clicking it while hovering will just select the arrow).

Additionally, there are a number of code improvements in this PR.
- Constants are now used for colors in javascript.
- scss files are all imported into `diagram.tsx`, which allows these classes to override `reactflow` defaults, so it was possible to drop a number of `!important`s from our css.
- Arrow specific css was moved from `diagram.scss` into `floating-edge.scss`.
- Some inline styling has been moved into scss.
- Logic for deleting edges has been moved from the diagram component to the root model.
- `test-utils.ts` was moved into the `utils` directory.

Finally, this fixes one bug that manifested in CLUE, but not elsewhere.
- In CLUE, it is possible for there to be multiple diagram instances, and some of them might be hidden. This was a problem when referencing arrowheads by id, because the first arrowhead might be invisible, making the arrowheads not appear in the visible diagram views. This PR fixes the problem by rendering arrowheads directly in the custom edges, rather than using the `marker-end` css property.

Big question (especially for @scytacki): I originally had the functions in `graph-utils.ts` as views on the variable model. However, I was having problems with the results not refreshing properly. (This was definitely the case for `validInputs` but I'm less certain about the other two functions.) They work in the current form, but we of course lose the benefit of MST caching the values, so they are less efficient than they could be. On the other hand, using this setup we can type the inputs as `VariableType`s without worrying about circular definitions. In any case, is the current set up adequate, or is there a better way to accomplish this?


src/diagram/components/icons/delete-button.tsx
```diff
@@ -0,0 +1,20 @@
...
+  onClick?: () => void;
+  x: number;
+  y: number;
+}
+export const IconDeleteButton = ({ onClick, x, y }: IProps) => {
+  const size = 20;
+  // This svg has been modified from the original from zepelin. Be careful when updating or replacing it.
+  return (
+    <svg width={`${size}px`} height={`${size}px`} viewBox={`0 0 ${size} ${size}`} x={x - size / 2} y={y - size / 2} xmlns="http://www.w3.org/2000/svg">
+      <g className="delete-button" id="M2Studio" onClick={onClick} stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
```
emcelroy, line 13: Should the `M2Studio` id value be removed or changed to something more generic?

src/diagram/components/connection-line.tsx
```diff
@@ -1,16 +1,33 @@
...
       <path
         className="react-flow__connection-path"
         data-testid="connection-line"
         fill="none"
-        stroke="#0081ff"
+        stroke={selectedBlue}
         strokeWidth={2}
-        d={`M${fromX},${fromY} C ${fromX} ${toY} ${fromX} ${toY} ${toX},${toY}`}
+        d={d[0]}
+        // d={`M${fromX},${fromY} C ${fromX} ${toY} ${fromX} ${toY} ${toX},${toY}`}
```
emcelroy, line 30: Did you mean to leave this commented-out line here?

emcelroy (approved):
This looks great to me 👍 I left a couple non-blocking comments about small things in the code.

When trying out the changes, I was able to repeatedly get to a state where the delete button for an arrow detached from the arrow. It seems like if an arrow forms a "simple curve" -- I mean where it's almost like a quadrant of a circle -- the button gets detached. See below.

image

Ideally that should be fixed, but I don't know how difficult it'd be. Everything else seems to work fine.

FWIW, it does seem to me that it'd be better to have the graph-utils.ts functions as views on the variable model, but I don't have a sense of how to fix the problems you were seeing with that. I'm sure Scott will have something more useful to say about it.

tealefristoe:
Thanks for the feedback @emcelroy .

I was a bit nervous that the position of delete buttons for single curved arrows would be a problem, but I actually figured out a solution that seems to work well pretty quickly. I was never able to get the arrows to appear from the left/right to the top/bottom, and I feel like the math might be slightly different in that case. Were you ever able to see this case?

emcelroy:
> I was a bit nervous that the position of delete buttons for single curved arrows would be a problem, but I actually figured out a solution that seems to work well pretty quickly. I was never able to get the arrows to appear from the left/right to the top/bottom, and I feel like the math might be slightly different in that case. Were you ever able to see this case?

The latest changes fix the problem for me 👍 

No, I never see an arrow coming from the left or right side of an output node attached to the top or bottom of an input node. As you move nodes around, an arrow's origin always switches to the top or bottom of the output node a little bit before the other end switches from the sides to the top or bottom of the input node. So it seems like you can't get into that state.

kswenson:
> Big question (especially for @scytacki): I originally had the functions in `graph-utils.ts` as views on the variable model. However, I was having problems with the results not refreshing properly. (This was definitely the case for `validInputs` but I'm less certain about the other two functions.) They work in the current form, but we of course lose the benefit of MST caching the values, so they are less efficient than they could be. On the other hand, using this setup we can type the inputs as `VariableType`s without worrying about circular definitions. In any case, is the current set up adequate, or is there a better way to accomplish this?

MST views caching only applies to getter functions, i.e. views without arguments (`get fooAccessor() { ... }`), so there's no loss where the other two functions are concerned. `validInputs()` should have worked as such a view on the `VariableModel`, however, so the fact that it wasn't behaving as you expected is concerning. What, exactly, was the symptom?

scytacki (approved):
This looks great to me. The Arrow component approach seems like a good solution.

With the amount of changes here, I don't think it is worth holding it up to figure out the `validInputs` issue.

tealefristoe:
@kswenson The symptom was the connect-ability not always updating when testing connections between different nodes. It's been long enough that I don't remember the exact situation, but it was something like nodes sometimes showing up as not connectable after establishing and breaking certain connections. However, I'm trying to recreate the problem now, and I'm having trouble doing it. So I'm going to make `validInputs` a get view and change it back to a regular function if bugs show up in production.

tealefristoe:
@kswenson @scytacki Just saw Scott's message about `validInputs`. I'll keep it as is but put a comment in about switching it to a view.

PR #82, opened 2023-06-12
185375603 Adjust reactflow viewport when the model's viewport changes.

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/185375603

This PR partially solves the issue of updating the `reactflow` viewport when the model changes (such as undoing a pan or zoom). The solution is only partial because I ran out of time working on it. Outstanding issues:
- Zoom in/zoom out do not properly undo/redo after the initial undo. It seems like the viewport change is not being correctly recorded in these cases.
- There are probably some user actions (like pinching to zoom) that are not being recorded, but I didn't even have time to look at these cases.

### DIFF
--- a/diagram-view/package-lock.json
+++ b/diagram-view/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@concord-consortium/diagram-view",
-  "version": "0.0.47",
+  "version": "0.0.48",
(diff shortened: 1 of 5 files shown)


emcelroy (approved):
Looks good to me 👍

PR #83, opened 2023-06-14
185298864 Simulator Tile Infrastructure

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/185298829

This work lays the groundwork for the Simulator tile in CLUE. This pretty much just involves exporting the `VariableSnapshot` type.

I also snuck in a number of improvements to the read only version of the diagram tile :)

### DIFF
--- a/src/diagram/models/dq-root.ts
+++ b/src/diagram/models/dq-root.ts
@@ -31,10 +31,10 @@ export const DQRoot = types.model("DQRoot", {
   }
 }))
 .views(self => ({
-  get reactFlowNodes() {
+  getReactFlowNodes(readOnly?: boolean) {
     const nodes: Node[] = [];
     self.nodes.forEach((node) => {
-      nodes.push(...node.getReactFlowNodes(self));
+      nodes.push(...node.getReactFlowNodes(self, readOnly));
     });
     return nodes;
   },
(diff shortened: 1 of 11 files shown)


emcelroy (approved):
Nice 👍

src/diagram/models/dq-root.ts
```diff
@@ -31,10 +31,10 @@ export const DQRoot = types.model("DQRoot", {
   }
 }))
 .views(self => ({
-  get reactFlowNodes() {
+  getReactFlowNodes(readOnly?: boolean) {
```
scytacki, line 34: @tealefristoe this will decrease the performance because MST won't be able to cache this response anymore. Since this call is probably done many times as nodes are dragged around, it is probably worth fixing this.

There might be a better way, but you could make two getters, one or read only one for not read only.

tealefristoe: Thanks for catching this. For setting up two getters, could I do something like this?

```getReactFlowNodes(readOnly?: boolean) { ... },
get reactFlowNodes() { return self.getReactFlowNodes(); },
get reactFlowNodesReadOnly() { return self.getReactFlowNodes(true); }```

Or do I need to duplicate the code in `get reactFlowNodes()` in two different getters, like below?

```get reactFlowNodes() {
  ...
  node.getReactFlowNodes(self)
  ...
}

get reactFlowNodesReadOnly() {
  ...
  node.getReactFlowNodes(self, true)
  ...
}```

kswenson: Either should work -- `get` accessor functions without arguments should be automatically cached independent of implementation.

PR #85, opened 2023-06-28
185425692 Extend Variables for Simulations

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/185425692

This PR expands `Variable` with new fields that help it support simulations in CLUE.


scytacki (approved):
This looks good to me.

PR #86, opened 2023-09-11
185952340 Variable Slider

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/185952340

This PR adds temporary values to variables, which only become permanent when explicitly set. It also adds a simple slider widget that can be used to change a variable's value. This slider is currently being used in the gripper simulator tile in CLUE, but might be used more generally for variables in the future.


scytacki (approved):
Looks good to me. 👍

PR #87, opened 2023-09-20
185952340 Fix Slider

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/185952340

This PR fixes a bug where undefined values were being saved to the variable when the user clicked on the slider handle but didn't move it.


bacalj (approved):
👍 looks good to me!

PR #90, opened 2024-01-17
ignore unused inputs when checking validity

bgoldowsky:
Fixes PT-186817666 

The bug report corresponded to a TODO in the file. Instead of bailing out any time an input has no value, first check whether that input is actually used.

### DIFF
--- a/src/diagram/models/variable.test.ts
+++ b/src/diagram/models/variable.test.ts
@@ -114,6 +114,38 @@ describe("Variable", () => {
     expect(variable.computedUnitIncludingMessageAndError).toEqual({});
   });
 
+  it("with 2 inputs one of which is valueless, result is valueless", () => {
+    const container = GenericContainer.create({
+      items: [
+        {id: "inputA", name: "a", value: 111.1},
+        {id: "inputB", name: "b"},
+        {id: "variable", value: 123.5, unit: "m", inputs: ["inputA", "inputB"], expression: "a+b"}
+      ]
+    });
+    const variable = container.items[2] as VariableType;
+
+    expect(variable.numberOfInputs).toBe(2);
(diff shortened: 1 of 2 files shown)


tealefristoe (approved):
Looks great 👍

bgoldowsky:
@tealefristoe @scytacki all the build process changes have been added to this PR.  It could use an additional review.

One remaining issue is that the `npm run build` process is creating extranous `.d.ts` files outside of the `dist` directory, I am not sure why.

tealefristoe (approved):
I won't pretend to understand all the changes to the build system here, but from what I could tell it looks good. -19750 lines... that's a number I like to see!

tejal-shah:
@bgoldowsky , is there any area that you'd recommend for regression testing? Since this is a huge PR, I'd appreciate it if you could give some pointers for focusing testing. I have checked the adding/inserting/editing variables directly and from their dialogs, adding/deleting connections, units and the behaviour of the output variables when adding connections which has the bug that this PR is trying to fix. Anything else you can think of?

bgoldowsky:
@tejal-shah Other than the fix to the PT issue linked in the description, this does not change any functional code, just the process by which the library is built and embedded into CLUE.  So my best guess would be to spot check a few things in each general area:
* Is the layout and design the same as before?
* Does general interactive functionality work?
* Does it load and save to the DB properly?
* Do toolbar functions work?
* Does interaction with other tiles work?
It sounds like you've probably done all of that.

tejal-shah:
@bgoldowsky , thank you, this is very helpful. I've now completed testing and we should be able to merge this in. I'm not hitting the merge button since I'm not sure if there's anything else remaining here or if someone else needs to approve it. Please feel free to merge as you see fit.

PR #93, opened 2024-10-07
don't register empty string units

scytacki:
- pluarlize returns an empty string singular for "S"
- in that case just register the initial unit ("S") instead of an aliased unit
- remove the try catch around creating the unit. This same try catch was not in place when the units were created in initializeMath, so this way we'll fail faster if there are more cases we haven't handled yet.
- unify the combination of addUnit and createUnit so the code is more clear.


bgoldowsky (approved):
LGTM 👍

PR #94, opened 2024-10-10
New rounding rule

scytacki:
PT-186820341

This branch was not merged. I'm not sure if it should have been?
It looks like it was the basis of diagram-view-0.0.54
However it seems like when diagram-view-1.0.0-beta.1 was released this change wasn't included?

Also from npm: https://www.npmjs.com/package/@concord-consortium/diagram-view?activeTab=versions
version 0.0.54 was never published.


scytacki:
Looks good to me. I can't officially review because I created the PR.

PR #95, opened 2024-11-04
better invalid unit error handling

scytacki:
- the UnitManager now doesn't crash when initializeMath is called and there is an invalid unit
- the `_` char is explicitly allowed to be a unit character
- invalid units are not added to the unit manager
- updated custom-math test so each test is isolated with its own instance of the units manager.
- removed what seemed like an unnecessary explicit "addUnit" call in variable.test.ts
- grouped invalid unit tests in variable.test.ts
- added tests for invalid characters of various types.


bgoldowsky (approved):
LGTM 👍

PR #98, opened 2026-05-20
Validate publish tag is semver before stamping

scytacki:
## Summary

Two small follow-ups to the tag-driven publish workflow added in #91, aligned with the equivalent workflow in [concord-consortium/accessibility-tools#3](https://github.com/concord-consortium/accessibility-tools/pull/3):

- **Validate the tag is semver before stamping.** The `diagram-view-v*` trigger glob matches any suffix (`diagram-view-v1.2`, `diagram-view-v1.2.3.4`, `diagram-view-v1.2.3-`, etc.), so a malformed tag would slip through and either fail mid-publish or push a broken version to npm. A regex guard now rejects non-semver tags up front with a clear `::error::` annotation.
- **Expand the comment in [`src/diagram/version.ts`](src/diagram/version.ts)** to explain why the workflow stamps the constant via `sed` rather than just doing `import pkg from "../../package.json"; export const version = pkg.version;` like accessibility-tools does. The short answer: this package is built with `tsc`, which preserves JSON imports as runtime `require`/`import` statements rather than inlining them — after compilation, the relative path would resolve against the `dist/` layout where `package.json` isn't present. Bundler-built siblings (webpack/rollup/tsup) inline JSON at build time and can use the import-from-package.json pattern safely.

The second change is comment-only; behavior is unchanged.

## Test plan

- [x] Push a malformed tag (e.g. `diagram-view-v1.2`) on a branch and confirm the workflow fails at the `Resolve version and dist-tag from git tag` step with the expected `::error::` message, before any stamping or publish happens.
- [x] Push a valid `diagram-view-v*-pre.N` tag and confirm the existing publish path still works end-to-end (validation passes, package.json is stamped, sed updates `src/diagram/version.ts`, `npm publish` succeeds).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish-library.yml
+++ b/.github/workflows/publish-library.yml
@@ -48,8 +48,8 @@ jobs:
         # X.Y.Z-prerelease.
         run: |
           VERSION="${GITHUB_REF_NAME#diagram-view-v}"
-          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.-]+)?$ ]]; then
-            echo "::error::Tag version '$VERSION' is not valid semver (expected X.Y.Z or X.Y.Z-prerelease)"
+          if ! npm version --no-git-tag-version --dry-run "$VERSION" >/dev/null 2>&1; then
+            echo "::error::Tag version '$VERSION' is not valid semver for npm version"
             exit 1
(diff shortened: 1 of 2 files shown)


dougmartin (approved):
Looks good 👍

